PR #3, opened 2024-07-11
Persist fw_version once installation got confirmed

tsagadar:
After a successful update, the fw_version should be set to the image name. This will prevent an endless loop of device updates.


jameshilliard:
> This will prevent an endless loop of device updates.

Yeah, we were using a config var to set the fw_version, wonder if it would be a good idea to extend the DDI API with a dedicated method of reporting version at some point.

PR #13, opened 2024-07-12
Support for multiple device types and more

tsagadar:
Main goal was to introduce `hw_model` and `hw_revison` to support different device types. Values are reported by devices (or left at "default") and gooseBit then choses the latest _matching_ firmware image (if set to "latest").


b-rowan:
Gonna merge for now, we can look at changing that timeout later.

PR #14, opened 2024-07-12
Remove `revision` refrences from URLs.

### LINKED ISSUE #11: Support for different device models/revisions
Firmware files encode the device model as the first part of the filename (and optionally its revision in the second part). This model/revision should be mapped with the model/revision reported by the device through the `configData` endpoint. This then allows to ensure that a newly registered device gets the firmware that matches its model/revision (as well as limiting the presented firmwares when manually assigning to a device).

This would require the config data (key/value pairs) or at least the model/revision to be persisted on the device (second approach seems simpler). If no model/revision key is present, gooseBit could assume "default" (for simple installations / quick getting started with a single device type).

b-rowan:
Removes the `revision` parameter from URLs to simplify slightly.  Close #11.


goosebit/updater/updates.py
```diff
@@ -60,7 +60,7 @@ def dl_endpoint(self):
         return "download_file"
 
     def generate_chunk(
-        self, request: Request, tenant: str, revision: str, dev_id: str
+        self, request: Request, tenant: str,  dev_id: str
```
easybe, line 63:
```suggestion
        self, request: Request, tenant: str, dev_id: str
```

easybe (approved):

goosebit/updater/updates.py
b-rowan: Good catch, forgot to install pre-commit hooks.

PR #16, opened 2024-07-12
Add install progress information

b-rowan:
- Add install progress to both device table and home table.
- Add progress bar to top of logs page.
- Add `progress` value to database.

### DIFF
--- a/goosebit/api/devices.py
+++ b/goosebit/api/devices.py
@@ -35,6 +35,7 @@ async def devices_get_all() -> list[dict]:
             "fw_file": device.fw_file,
             "hw_model": device.hw_model,
             "hw_revision": device.hw_revision,
+            "download_percentage": device.download_percentage,
             "state": device.last_state,
             "force_update": manager.force_update,
             "last_ip": device.last_ip,
--- a/goosebit/api/devices.py
+++ b/goosebit/api/devices.py
@@ -35,7 +35,7 @@ async def devices_get_all() -> list[dict]:
             "fw_file": device.fw_file,
             "hw_model": device.hw_model,
             "hw_revision": device.hw_revision,
-            "download_percentage": device.download_percentage,
+            "progress": device.progress,
             "state": device.last_state,
             "force_update": manager.force_update,
             "last_ip": device.last_ip,
--- a/goosebit/models.py
+++ b/goosebit/models.py
@@ -14,7 +14,7 @@ class Device(Model):
     hw_model = fields.CharField(max_length=255, null=True, default="default")
     hw_revision = fields.CharField(max_length=255, null=True, default="default")
     last_state = fields.CharField(max_length=255, null=True, default="unknown")
-    download_percentage = fields.IntField(null=True)
+    progress = fields.IntField(null=True)
     last_log = fields.TextField(null=True)
     last_seen = fields.BigIntField(null=True)
     last_ip = fields.CharField(max_length=15, null=True)
--- a/goosebit/models.py
+++ b/goosebit/models.py
@@ -14,6 +14,7 @@ class Device(Model):
     hw_model = fields.CharField(max_length=255, null=True, default="default")
     hw_revision = fields.CharField(max_length=255, null=True, default="default")
     last_state = fields.CharField(max_length=255, null=True, default="unknown")
+    download_percentage = fields.IntField(null=True)
     last_log = fields.TextField(null=True)
     last_seen = fields.BigIntField(null=True)
     last_ip = fields.CharField(max_length=15, null=True)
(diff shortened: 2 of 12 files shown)


tsagadar:
Two issues I see with this PR

1. Relies on arbitrary swupdate log data. So it never will work with a different client and format could easily change between swupdate versions. DDI does have a progress indication in the deployment feedback but I could not find a documentation explaining how it should be used. swupdate uses it - but not really for download progress (have not fully understood the code yet).
2. Database writes should be kept at a minimum if solution should scale to many devices without requiring a horizontal scale out. So maybe just keep that information in memory?

How critical is it to have this feature?

b-rowan:
I don't think it's super critical, but it is a pretty useful feature.

> Relies on arbitrary swupdate log data. So it never will work with a different client and format could easily change between swupdate versions. DDI does have a progress indication in the deployment feedback but I could not find a documentation explaining how it should be used. swupdate uses it - but not really for download progress (have not fully understood the code yet).

This could just be implemented directly on the DDI endpoint, although it definitely seems like it's not documented clearly.  I think it's simple enough to remove in the future, if it is the case that something is wrong with it, but this seems like a good enough intermediate solution.

> Database writes should be kept at a minimum if solution should scale to many devices without requiring a horizontal scale out. So maybe just keep that information in memory?

I'm totally fine with keeping it in memory, but in that case, the log should also be kept in memory.  They are both being written at the same time with this `device.save()`.

https://github.com/UpstreamDataInc/goosebit/blob/1c52ee1a83f2563005fedf5540434d5799ab8a80/goosebit/updater/manager.py#L192

b-rowan:
> Relies on arbitrary swupdate log data. So it never will work with a different client and format could easily change between swupdate versions. DDI does have a progress indication in the deployment feedback but I could not find a documentation explaining how it should be used. swupdate uses it - but not really for download progress (have not fully understood the code yet).

Another thing here, the regex I'm using isn't very specific.  Would making it more specific be better?  Technically the line is `[channel_callback_xferinfo] : Downloaded 14% (31741 of 226675 kB).`, would this work better with `r'[channel_callback_xferinfo] : Downloaded (\d+)%'` instead?

b-rowan:
> DDI does have a progress indication in the deployment feedback but I could not find a documentation explaining how it should be used. 

You mean the `/feedback` endpoint with `status.result.progress` I assume?  In our case, that reports `cnt = 0-1` and `of = 1`, which is basically useless...

tsagadar:
While implementing a "rollout" feature (not yet ready for PR) I also realized that `poll_time_updating` was likely introduced to have frequent updates on the `/feedback` endpoint. By itself, swupdate calls `/feedback` 6x (in our case) - but mostly at the beginning and at the end of the process - so download cannot be nicely tracked.

Again would prefer to keep load as low as possible - so the short polling time during the updates is somewhat problematic.

tsagadar:
In general: I love the capability to see the update logs (was not aware that this is possible) but would not go overboard to see the download progress (usually is like 20s in our case).

tsagadar:
> Another thing here, the regex I'm using isn't very specific. Would making it more specific be better? Technically the line is [channel_callback_xferinfo] : Downloaded 14% (31741 of 226675 kB)., would this work better with r'[channel_callback_xferinfo] : Downloaded (\d+)%' instead?

If the feature is kept, I'd prefer the current (less specific version). This makes it a bit more robust and don't think it likely that it suddenly would match something unwanted.

b-rowan:
> While implementing a "rollout" feature (not yet ready for PR) I also realized that `poll_time_updating` was likely introduced to have frequent updates on the `/feedback` endpoint. By itself, swupdate calls `/feedback` 6x (in our case) - but mostly at the beginning and at the end of the process - so download cannot be nicely tracked.
> 
> Again would prefer to keep load as low as possible - so the short polling time during the updates is somewhat problematic.

This is precisely why `poll_time_updating` is configurable.  We use updates every 5s when updating, which makes the progress a lot more precise, especially it a smaller scale.  At larger scale, you can just tone back all of those settings, which makes it a bit less useful, but can still give you a nice idea of which devices are almost done by sorting in the table.  It's also a nice overview to glance at in the logs page.

b-rowan:
> In general: I love the capability to see the update logs (was not aware that this is possible) but would not go overboard to see the download progress (usually is like 20s in our case).

Fair enough, but since the progress is being sourced from data we already have I don't really see an issue with it?  Basically I don't see a case where this is every really a big negative, maybe the regex parsing is a bit inefficient and could cause slowdowns?

jameshilliard:
> In our case, that reports `cnt = 0-1` and `of = 1`, which is basically useless...

Well this tracks the update stage I think? Might not be entirely useless but def not that precise.

> Fair enough, but since the progress is being sourced from data we already have I don't really see an issue with it?

Note that tracking progress percentage with that regex is not an overall install percentage, it's only the download percentage so it's somewhat stage specific(although in our case it's usually the slowest operation). I think we want to track update stage and then percentage within each stage.

Issue #18, opened 2024-07-16
Roadmap Discussion


tsagadar:
Here some thoughts about potential improvements in gooseBit from my perspective. Would be nice to collect thoughts / ideas to see where the project should be going. Note: I will be on vacations for a few days.

### Features

1. UI support for rollouts
2. Rollout abort: automatically pause rollout when error rate is above a certain threshold.
3. CDN support: improve scalability by devices not directly downloading images from gooseBit server.
4. Prevent unnecessary updates after initial registration: support devices to report currently installed firmware version. Use to detect if update is required.

### Stability reviews / improvements

1. Switch "version" from date-time to Semantic Versioning (https://semver.org/). This allows ordering as well and seems more common.
2. Review use of "action_id": currently hard-coded to 1 - so not useful to match feedbacks to the originating action.
3. Review memory management: weakly link DeviceUpdateManager instances to prevent running out of memory.
4. Review database usage: minimize load that a single device poll request generates.
5. Introduce automated tests to validate core update logic.

### Operations Improvements

1. Logging
2. (Re-)introduce database migrations
3. Metrics: ideally using Open Telemetry
4. Validate that gooseBit works with Postgres

b-rowan:
Couple PRs opened to do a few of these, B1 and partial A3.

b-rowan:
Making this into a checklist, so we can handle each of these independently - 

## Features:
 - [ ] UI support for rollouts
 - [ ] Rollout abort
 - [ ] CDN support for update files
 - [ ] Prevent unnecessary initial updates
 
 ##  Stability Improvements:
 - [x] Allow semantic versioning
 - [ ] Review `action_id` usage
 - [ ] Improve memory management of `UpdateManager`
 - [ ] Improve poll performance by reducing DB load
 - [ ] Introduce automated testing
 
 ## Operational Improvements:
 - [ ] Add logging
 - [ ] Database migrations
 - [ ] Add metrics
 - [ ] Add postgres support

PR #19, opened 2024-07-19
Alternate update location support

b-rowan:
Add support for alternative update locations (such as CDN as mentioned in #18)


b-rowan:
I frankly have no idea how well this will work for the CDN support, this is mostly an attempt at adding ANY kind of support for modifying the base file directory.  It doesn't handle creating directories, and will just raise attribute errors, and also can't iterate the remote directory to list files, but I may need some assistance on that side of things.  Not sure what the best way to iterate a remote CDN directory is, we may need to specify that if using a remote server for firmware files, there should be at least a `/list` or `/firmware` directory, so that the firmware can be accurately listed and selected from the UI.  let me know what you think there.

@tsagadar I also added you as a maintainer so that I can have you review these types of PRs.

easybe:
As I am currently on vacation, my reviewing capabilities are rather limited.

What I can say is, for our CDN we need a way to download the artifacts from an URL that does not contain the controller ID. That can be the API with e.g. an access token. Then we need a way to specify the URL (pointing to the CDN) that is passed to the client.

easybe:
Alternatively, uploading the artifacts directly to S3 and merely referencing them in goosBit might als be an option.

tsagadar:
Usually, a CDN is not a permanent store of files, but rather a cache close to the consumer. So the CDN still needs to be configured with an origin that tells it where to download files from (in our case: gooseBit). So two things would be important for CDN support:

1. gooseBit should not hand-out URLs that point to itself, but rather ones that point to the CDN
2. The URLs handed-out to the devices should not contain a device-specific identifier. If they do, then the CDN would always have a cache miss and the same amount of requests would hit gooseBit.

How gooseBit stores the firmware files - on the local filesystem, a remote filesystem or even the database is a different issue.

easybe:
But, in our case, we could simple upload the artifacts to S3 (as we already do) and point the CDN there. Therefore, we only need to be able to set an arbitrary URL instead of uploading a file. This way we can keep things simple and also make our deployment cleaner as we do not need to upload large files through the API.

goosebit/uri.py
```diff
@@ -0,0 +1,44 @@
+import copy
+import dataclasses
+import urllib.parse
+from pathlib import Path
+
+
+@dataclasses.dataclass
+class RemoteFirmwareURI:
```
easybe, line 8: Instead of all of this I would introduce a `Firmware` DB model with at least a `name`, `version` and `uri` (can be local or remote) field. In Django you can also add methods to Models if needed, don’t know about this ORM. AFAICT the model would replace `FirmwareArtifact` and return either an URL to a local or a remote file depending on the value of the `uri` model field.

easybe (changes requested):

easybe:
Hi,

> On 21 Jul 2024, at 15:46, Stefano Babic ***@***.***> wrote:
> 
> As you mentioned here, this raises to me the curiosity to know why
> Django was not chosen (it will be my first choice). It is just for my
> curiosity.

I guess a more modern (async) framework/libraries were chosen.

> 
> I think that the best way is that the data are automatically detected
> from the SWU. this requires some kind of inspection, but it is not so
> hard. A SWU is just a cpio archive, and extracting sw-description from
> the SWU is easy. It is mandatory in SWU to add a "version" attribute.
> gooseBit could then extract the version from the SWu (version
> attribute), and then the filename could even be JaneDoe.swu, it does not
> matter. And gooseBit can check if the same version is uploaded with
> different names by checking version and computing a hash on the whole SWU.
> 
> Going further, model and revision could be read from the
> hardware-compatibility attribute. It is just to understand how goposeBit
> should know about the SWU details and extract data from sw-description.

I like the Idea. We already briefly touched on the topic but figured it is too complicated. But in that case, we should give it a try.

Cheers,
Ezra

goosebit/uri.py
b-rowan: This is a good idea, let me see what I can do.  Should be no problem adding additional functionality to the model, it is represented as a class like any other ORM.

b-rowan:
> I guess a more modern (async) framework/libraries were chosen.

I'm not super familiar with django in general, I find FastAPI to be very easy to work with, and it is also extremely fast.  There are also some nice-to-have features, like the automatic swagger API documentation for users who want to wrap that.

 >> Going further, model and revision could be read from the
 hardware-compatibility attribute. It is just to understand how goposeBit
 should know about the SWU details and extract data from sw-description.
> I like the Idea. We already briefly touched on the topic but figured it is too complicated. But in that case, we should give it a try.

I'm not sure familiar with how that would be parsed, especially for remote firmware files, something to think about.  Locally that should be no issue.

goosebit/uri.py
b-rowan: I think the UI should also probably change based on whether gooseBit is configured for local or remote file handling, then you can just enter a URL to the file or upload a file.

b-rowan:
> What I can say is, for our CDN we need a way to download the artifacts from an URL that does not contain the controller ID. That can be e.g. a management API with an access token. Then we need a way to specify the URL (pointing to the CDN) that is passed to the client.

I'm not actually entirely sure why the device ID is included in the URL for downloads, I'll remove that.

easybe:
> I'm not actually entirely sure why the device ID is included in the URL for downloads, I'll remove that.

That is how DDI/hawkBit does it, URLs are controller-specific. I would leave it as the default behavior. For us it is enough, if we can set custom remote URLs.

b-rowan:
> That is how DDI/hawkBit does it, download URLs are controller-specific. I would leave it as the default behavior. For us it is enough, if we can set custom remote URLs.

I think it actually makes more sense to just have it redirect to an arbitrary download endpoint, I see no reason to duplicate the hawkbit/DDI way of doing things in this case, since even in our case with having cloudflare between the devices and the updater it should improve download speed vs having the controller specific downloads.  By the time the device has requested the location of the firmware download, we have enough information to construct the correct download endpoint for that device without needing to use the device id or tenant during the download.

b-rowan:
Closing in favor of #30

PR #21, opened 2024-07-20
Improve session handling

b-rowan:
Improve session token handling by parsing JWTs properly.


b-rowan:
Also adding a swap from `python-jose` to `joserfc`, as `python-jose` has 2 CVEs active and seems to no longer be maintained (last updated in 2021).

b-rowan:
The specific reasoning for this PR is that if we decide to make `SECRET` a environment variable or setting in the future, any case where it is the same across restarts, the same `session_id` value for a user will be valid across restarts, which was not the case prior to this because the `sessions` dict would get cleared on each restart.

PR #24, opened 2024-07-22
Add metrics and Prometheus support

b-rowan:
Adds some very basic metrics (registered users and device count), along with the default metrics provided by OpenTelemetry and its FastAPI integration.  We may want to look at making the different exporters into feature flags.

This should provide a very basic starting point to start adding better telemetry, along with spans and logging.


b-rowan:
Closes #26.

tsagadar (approved):
Worked with my local Prometheus installation. Great to have this in place.

PR #30, opened 2024-07-25
Add remote file handling

### LINKED ISSUE #28: CDN support for update files

b-rowan:
Adds remote firmware handling for files.  Based on #29.  Closes #28. Closes #29.

Adding independently so they can be merged independently if there are concerns on this PR.

Main concern here is that I don't have a better way to get the hash of the upgrade file (required for swupdate artifact) without just downloading the file when the link is added.


goosebit/models.py
```diff
@@ -38,3 +45,65 @@ class Rollout(Model):
...
     failure_count = fields.IntField(default=0)
+
+
+class FirmwareCompatibility(Model):
+    id = fields.IntField(primary_key=True)
+    hw_model = fields.CharField(max_length=255)
+    hw_revision = fields.CharField(max_length=255)
+
+
+class FirmwareUpdate(Model):
```
easybe, line 56: I would just call it `Firmware` as it does not have to be an update, it could technically also be a downgrade (if we would start to allow that) or the initial image.

easybe (commented):
IMHO downloading the artifact when creating a new DB entry to get the hashes is a reasonable approach.

I’ve only skimmed over the changes as reviewing on the phone is not that fun.

easybe:
Oh, and please note this comment: https://github.com/UpstreamDataInc/goosebit/commit/c5531765c7e98009e76a81d4d656c04370653561#r144602289

tsagadar:
I think we should slow down somewhat: as there are no automatic test in the code base, every PR breaks some functionality. Manual testing (what I was doing so far) takes forever. So we better ensure to have tests for the existing functionality before adding / refactoring code.

I created automated API integration tests (focus on the DDI) yesterday - and currently adjust the test code to match this PR. Ideally I could finish these tests (few of them will fail as the code is broken), integrate them into this PR and ensure that everything passes before merging.

goosebit/models.py
tsagadar: The rollout model needs to be adjusted as well - instead of hw_model, hw_revision, fw_file it only should reference a firmware.

This then also means work in Manager.get_rollout to query the correct firmware.

goosebit/models.py
```diff
@@ -38,3 +45,65 @@ class Rollout(Model):
...
+    hash = fields.CharField(max_length=255)
+    version = fields.CharField(max_length=255)
+    compatibility = fields.ManyToManyField(
+        "models.FirmwareCompatibility",
+        related_name="updates",
+        through="update_compatibility",
+    )
+
+    @classmethod
+    async def latest(cls):
```
tsagadar, line 69: This needs a parameter "device" to properly filter for the latest firmware.

goosebit/updater/manager.py
```diff
@@ -170,16 +169,26 @@ async def get_update_file(self) -> FirmwareArtifact:
...
+        return await FirmwareUpdate.get(
+            uri=UPDATES_DIR.joinpath(device.fw_file).absolute().as_uri(),
+            compatibility=compat,
+        )
 
     async def get_update_mode(self) -> str:
         device = await self.get_device()
 
         file = await self.get_update_file()
-        if file.is_empty():
```
tsagadar, line 179: Why removing this case? At least when the system starts and devices connect there are no firmware available and no update should be rolled out.

goosebit/updater/manager.py
```diff
@@ -170,16 +169,26 @@ async def get_update_file(self) -> FirmwareArtifact:
...
+        )
+        return await FirmwareUpdate.get(
+            uri=UPDATES_DIR.joinpath(device.fw_file).absolute().as_uri(),
+            compatibility=compat,
+        )
 
     async def get_update_mode(self) -> str:
         device = await self.get_device()
 
         file = await self.get_update_file()
```
tsagadar, line 190: Rename file / get_update_file to firmware / get_firmware (or similar)

goosebit/updater/manager.py
```diff
@@ -170,16 +169,26 @@ async def get_update_file(self) -> FirmwareArtifact:
...
-        if file.is_empty():
-            mode = "skip"
-            self.poll_time = POLL_TIME
-        elif file.name == device.fw_version and not self.force_update:
+        if file.path.name.split(".")[0] == device.fw_version and not self.force_update:
```
tsagadar, line 191: Version is explicitly stored on firmware, so no need parsing the filename.

.pre-commit-config.yaml
```diff
@@ -14,12 +14,12 @@ repos:
...
-    hooks:
-    -   id: pytest
-        name: pytest
-        entry: poetry run pytest tests
-        language: system
-        types: [ python ]
-        pass_filenames: false
-        always_run: true
\ No newline at end of file
+#-   repo: local
```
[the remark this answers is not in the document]
tsagadar: Why remove this in the first place? Should really have tests in the system.

goosebit/api/download.py
```diff
@@ -1,20 +1,29 @@
...
-@router.get(
-    "/{file}",
-    dependencies=[
-        Security(validate_user_permissions, scopes=[Permissions.FIRMWARE.READ])
-    ],
-)
-async def download_file(request: Request, file: str):
-    filename = UPDATES_DIR.joinpath(file)
-    return FileResponse(filename, media_type="application/octet-stream")
+@router.get("/by_name/{filename}")
```
tsagadar, line 11: What will this endpoint be used for?

goosebit/db.py
```diff
@@ -22,6 +23,10 @@ async def init():
     await command.migrate()
     await command.upgrade(run_in_transaction=True)
     await Tortoise.generate_schemas(safe=True)
+    for update in await FirmwareUpdate.all():
```
tsagadar, line 26: Instead of `all()` better to filter on db using `filter()`.

goosebit/models.py
```diff
@@ -38,3 +45,65 @@ class Rollout(Model):
...
+
+    @property
+    def path(self):
+        return Path(url2pathname(unquote(urlparse(self.uri).path)))
+
+    @property
+    def local(self):
+        return urlparse(self.uri).scheme == "file"
+
+    def generate_chunk(self, request: Request) -> list:
```
tsagadar, line 85: Starting to overload the model classes? Looks like an API related aspect.

tsagadar (changes requested):
Few things I saw while writing tests. Some if them cause my current tests to fail.

tsagadar (commented):
I finished a first set of integration test - they can be found at https://github.com/husqvarnagroup/goosebit/tree/mm/add_simple_ddi_integration_tests

Besides the commit adding the tests there is an additional one that makes the tests pass. Most of the things are also annotated in this PR.

Note: `poetry run pytest` does not pass all the tests. Single execution of tests work (I do it using PyCharm). Will look into this soon (but not today)

b-rowan:
I'll try to take a look at this later tonight, at a conference this week so I don't have a ton of time, but I'll fix what I can ASAP.

> I think we should slow down somewhat: as there are no automatic test in the code base, every PR breaks some functionality. 

Agreed, but I really want to get the important core functionality we want figured out ASAP.  I'm not a big tests guy personally, I think it's definitely a good idea in this case, I just struggle with the differences between trying to finish important functionality and stabilizing the code/API.  There seem to be a lot of cases where the dev API needs to change for important things to happen, so I wasn't sure if we wanted to wait to write tests until after that.

goosebit/api/download.py
b-rowan: Fair enough, I think this can safely be removed, as all firmware can now just be queried by ID, even in the `/firmware/all` endpoint it reports the ID.

goosebit/db.py
b-rowan: Not sure what to filter by here, I suppose maybe `uri` starts with `file`?

goosebit/updates/swdesc.py
```diff
@@ -0,0 +1,52 @@
...
+            return None
+
+        swdesc = libconf.loads((await f.read(size)).decode("utf-8"))
+
+    swdesc_attrs = {}
+    try:
+        swdesc_attrs["version"] = semver.Version.parse(swdesc["software"]["version"])
+        swdesc_attrs["size"] = file.stat().st_size
+        swdesc_attrs["hash"] = sha1_hash_file(file)
+        swdesc_attrs["compatibility"] = [
```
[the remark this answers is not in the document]
b-rowan: Is there a standard on how this is supposed to be set?  We use `loadsync.v1` here, but I'm not sure if there is any standard here...

goosebit/updater/manager.py
```diff
@@ -170,16 +169,26 @@ async def get_update_file(self) -> FirmwareArtifact:
             if rollout and not rollout.paused:
                 file = rollout.fw_file
 
-        return FirmwareArtifact(file, device.hw_model, device.hw_revision)
+        if file == "latest":
+            return await FirmwareUpdate.latest()
+        if file == "pinned":
+            return await FirmwareUpdate.get(
+                uri=UPDATES_DIR.joinpath(device.fw_file).absolute().as_uri()
```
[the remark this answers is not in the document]
b-rowan: Need to take a look at this, not sure how the best way to have it still generate the artifact information and have it just skip with no file.  All the artifact generation being done in the firmware file makes things a bit tough, maybe I should split that method out and have it take `Firmware | None`?

goosebit/models.py
b-rowan: Moved to take `FirmwareUpdate` as a parameter.

goosebit/models.py
b-rowan: > The rollout model needs to be adjusted as well - instead of hw_model, hw_revision, fw_file it only should reference a firmware.
> 
> This then also means work in Manager.get_rollout to query the correct firmware.

This can get really tricky, if I link `Firmware` directly then all the stuff that wants to check for rollout that uses the device as a backup need to return `Firmware`, which means `fw_file` on `Device` needs to be replaced with `Firmware`, but then that gets rid of `latest` and `pinned`, which can make things pretty complicated.  Let me know what you guys think here.

b-rowan:
> I finished a first set of integration test - they can be found at https://github.com/husqvarnagroup/goosebit/tree/mm/add_simple_ddi_integration_tests
> 
> Besides the commit adding the tests there is an additional one that makes the tests pass. Most of the things are also annotated in this PR.
> 
> Note: `poetry run pytest` does not pass all the tests. Single execution of tests work (I do it using PyCharm). Will look into this soon (but not today)

Implemented these, and fixed your issue with pytest not working properly.  Just need to add `__init__.py` files to all directories leading to the tests to make them identify as modules.

b-rowan:
Mostly everything fixed for now, just a couple things I want to clarify on before I change anything else, check comments on unresolved when you get a chance @tsagadar @easybe.  Appreciate the pointers, this was a pretty large PR, but I think a pretty important one to deal with early on.

goosebit/models.py
tsagadar: Yes, it triggers a few follow-up changes. I have a branch where I introduced `update_mode` on Device (enum with `latest`, `rollout`, `assigned`, and `pinned`), so that `fw_file` could be changed to `firmware` on Device.

Will re-implement those changes on top of this branch to see if it makes sense.

goosebit/db.py
tsagadar: Thought about using `await Firmware.filter(url__startswith='file').all()`. Debatable if a good idea, as logic of `local` is the implemented in two different places. On the upside: with a lot of remote firmware it will be very fast.

goosebit/updates/swdesc.py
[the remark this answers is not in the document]
tsagadar: This is not my area of expertise. Maybe @easybe can provide information once he is back.

tsagadar (approved):

tsagadar:
I think this PR is ready to be merged - great stuff. Follow-up work can be done in a new branch / PR. Will look replacing fw_file with firmware.

PR #39, opened 2024-07-30
refactor: Move settings files into settings directory to improve organization

b-rowan:
Slight organization improvement by moving application settings related files away from files that should not be configured by a user.  Closes #42.


tsagadar:
Don't have a strong opinion here. Usually one would see (a single) settings file(s) at the top level IMHO. Don't we expose settings so that the user could change them?

b-rowan:
> Don't have a strong opinion here. Usually one would see (a single) settings file(s) at the top level IMHO. Don't we expose settings so that the user could change them?

Yes, this is actually the idea.  I'm not opposed to combining settings files, but even then, it's hard to differentiate between project control files and user defined settings.  The goal of this PR was to make it a bit simpler to understand, if only temporarily.

b-rowan:
Ill make a separate PR that we can compare, which will combine the files.

jameshilliard:
We should probably also allow passing the settings file path via an argument.

PR #41, opened 2024-08-01
Allow feed/flavor assignment in devices view & cleanups/fixes

tsagadar:
This PR should wrap up the initial rollout functionality: devices can be assigned to feeds/flavors, and the rollout selection got fixed. UI got cleaned-up to make it easier to use.

Also contains cleanups and bug fixes.

The one thing missing around rollouts is the automatic abort (pause) which will follow in a later PR.


b-rowan (approved):
Just a small question about this - 
```javascript
        stateLoadParams: (settings, data) => {
            // if save state is older than last breaking code change...
            if (data.time <= 1722434386000) {
                // ... delete it
                for (const key of Object.keys(data)) {
                    delete data[key];
                }
            }
        },
```

This is just a form of cache busting right?  Keeps the table state from becoming invalid?

Do we maybe want to move the timestamp from there to a constant stored in another js file then included, which could then be referenced?  Something like `LAST_TABLE_BREAKING_CHANGE`?

PR #42, opened 2024-08-01
refactor: combine settings files

b-rowan:
An alternative to #39 which combines logging and settings configuration into `settings.yaml`.  Closes #39.


tsagadar:
Personally I prefer this version: all in one place.

The file could be split (by comments) into two parts to show configs that should be reviewed for each installation and more internal settings that usually are ok to leave as is. Here an example (note: permissions of "ops" user won't work - seem to have a bug)

```
## Settings to adjust for each installation

# Database to be used
db_location: db.sqlite3

# Frequency that devices should check for available updates.
poll_time_default: 00:01:00

# Frequency to log information during an update. Defines how often download progress
# and device logs get updated.
poll_time_updating: 00:00:05

# User account for the frontend. Available permissions:
# "firmware.read", "firmware.write", "firmware.delete"
# "devices.read", "devices.write", "devices.delete"
# "rollouts.read", "rollouts.write", "rollouts.delete"
# "home.read"
users:
  - email: admin@goosebit.local
    password: admin
    permissions:
      - "*"
  - email: ops@goosebit.local
    password: ops
    permissions:
      - "home.read"

## Internal settings that usually don't need to be modified
poll_time_registration: 00:00:10
tenant: DEFAULT
metrics:
  prometheus:
    port: 9090
logging:
  version: 1
  formatters:
    simple:
      format: "%(asctime)s - %(name)s - %(levelname)s - %(message)s"
  handlers:
    console:
      class: logging.StreamHandler
      formatter: simple
      level: DEBUG
  loggers:
    tortoise:
      handlers: [ console ]
      level: WARNING
      propagate: yes
    aiosqlite:
      handlers: [ console ]
      level: WARNING
      propagate: yes
    multipart:
      handlers: [ console ]
      level: INFO
      propagate: yes
  root:
    level: INFO
    handlers: [ console ]
```

b-rowan:
> (note: permissions of "ops" user won't work - seem to have a bug)

Let me look into this.  I'll add comments as suggested.

tsagadar (approved):

PR #47, opened 2024-08-06
feature: create plugins architecture

### LINKED ISSUE #15: gooseBit plugin/module system design
There's a number of use cases where users of gooseBit may need to extend the updater application with custom functionality that for various reasons should be maintained separately from the upstream gooseBit project, ideally we want it to be possible to extend gooseBit without having to fork and modify the base project codebase.

For example we want to be able to implement a configuration management extension for provisioning of device keys as a gooseBit plugin/module.

In terms of how plugin configuration would look I'm thinking something like this via the yaml settings file may make sense:
```yaml
plugins:
  upstreamdata.cfd.provisioner:
    configkey1: configval1
    configkey2: configval2
    configkey3:
      subkey1: subval1
      subkey2: subval2
```

The idea would be that we have a python class/package with the import name `upstream.cfd.provisioner` that extends some sort of base module class from gooseBit which would be passed the decoded plugin child yaml configuration structure(a multilevel dictionary in the above example) when the module `__init__` method is called during plugin load after dynamically importing the `upstream.cfd.provisioner` python module class.

The module class implementation from the plugin project would then be expected to provide implementations for the various hook methods/variables that gooseBit would call in various places such as when the device sends config vars or when device poll various endpoints.

The module system would also likely need a way for plugins to define their own database models/tables which may need to reference the existing gooseBit models so would likely need orm/migration hooks for integrating that as well. Likely we want these plugin defined database models/tables to have some sort of namespacing pattern that ensures they won't conflict with other plugins or the base gooseBit project models/tables.

Some potentially relevant info:
https://github.com/team23/fastapi-module-loader
https://github.com/Renumics/spotlight/blob/main/renumics/spotlight/plugin_loader.py
https://fastapi.tiangolo.com/tutorial/dependencies/#fastapi-plug-ins
https://fastapi.tiangolo.com/features/#unlimited-plug-ins
https://packaging.python.org/en/latest/guides/creating-and-discovering-plugins/

b-rowan:
This PR aims to create a way to handle plugins in gooseBit.  Closes #15.


b-rowan:
@jameshilliard can you take a look at this and make sure it makes sense?  I think this should be simple enough, just allow plugins to define entry points, but just want to verify this is reasonable before I continue.

examples/example_plugin/README.md
```diff
@@ -0,0 +1,21 @@
+# gooseBit Example Plugin
+
+This is a basic example plugin to provide a framework for creating custom plugins for gooseBit.
+
+## Plugin Requirements
+
+Plugins are identified automatically with entry points. This means that you must define the relevant section in your `pyproject.toml` file.
+
```
jameshilliard, line 8: I'm thinking we should define plugins in the config file and dynamically load them based off of the import path in the config file.

b-rowan: The standard for this seems to be just that you can install them and they will automatically get loaded.  As long as  users are properly keeping track of dependencies this wont be a problem.

You can test this by running `poetry install --with examples` to load the plugin, and `poetry install --without examples` to unload it.

jameshilliard: > The standard for this seems to be just that you can install them and they will automatically get loaded.

I'm not that big a fan of auto-loading just based on what's installed, IMO they should be explicitly enabled in a config file as well so that merely installing plugins to the python environment doesn't actually change the behavior of the application(which would be somewhat unexpected behavior).

> As long as users are properly keeping track of dependencies this wont be a problem.

Yeah...so I don't think we want to change behavior based on installed deps in this way, I mean if a user installs a plugin but then decides not to use it they should be able to just not have it configured for it to not be loaded or something like that.

b-rowan: Ok, so we need another option in settings for `plugins`, which can be a list of plugin names.  Shouldn't be too hard.

jameshilliard: I think each plugin should be a configurable class(which can import various hooks) and the plugin name should be the import path.

jameshilliard: And configure like this maybe:
```yaml
plugins:
  upstreamdata.cfd.provisioner:
    configkey1: configval1
    configkey2: configval2
    configkey3:
      subkey1: subval1
      subkey2: subval2
```

goosebit/plugins/__init__.py
```diff
@@ -0,0 +1,10 @@
+from importlib.metadata import entry_points
+
+from goosebit.settings import PLUGINS
+
+entries = entry_points(group="goosebit.plugins")
+
+plugins = []
+
+for p in PLUGINS:
+    plugins.extend(entries.select(name=p))
```
jameshilliard, line 10: Maybe iterate plugins from configuration file by key, passing the plugin values as settings to plugin class `__init__`.

I think you can probably use something [import_module](https://docs.python.org/3/library/importlib.html#importlib.import_module) based like [these functions](https://github.com/django/django/blob/stable/5.1.x/django/utils/module_loading.py#L8-L35) for module loading from settings defined string keys.

b-rowan: When PR lol

b-rowan:
There's a couple things I like here that aren't super related to plugins, I will try to break them out.

PR #49, opened 2024-08-07
Prevent unnecessary initial updates / UI Refinements / Fixes

tsagadar:
Covers two more issues from the roadmap, additional UI improvements and bug fixes.

### DIFF
--- a/goosebit/api/firmware.py
+++ b/goosebit/api/firmware.py
@@ -25,13 +25,15 @@ async def firmware_get_all(
         return Q(uri__icontains=search_value) | Q(version__icontains=search_value)
 
     async def parse(f):
+        compatibility = await f.compatibility.all().values()
+
         return {
             "id": f.id,
             "name": f.path.name,
             "size": f.size,
             "hash": f.hash,
             "version": f.version,
-            "compatibility": list(await f.compatibility.all().values()),
+            "compatibility": compatibility,
         }
 
     total_records = await Firmware.all().count()
--- a/goosebit/ui/templates/rollouts.html
+++ b/goosebit/ui/templates/rollouts.html
@@ -34,41 +34,67 @@
                     aria-label="Close"
                 ></button>
             </div>
-            <div class="modal-body">
-                <input
-                    id="rollout-selected-name"
-                    class="form-control mb-3"
-                    placeholder="Name"
-                />
-                <input
-                    id="rollout-selected-feed"
-                    class="form-control mb-3"
-                    placeholder="Feed"
-                />
-                <input
-                    id="rollout-selected-flavor"
-                    class="form-control mb-3"
-                    placeholder="Flavor"
-                />
-                <select class="form-select" id="selected-fw"></select>
-            </div>
-            <div class="modal-footer">
-                <button
-                    type="button"
-                    class="btn btn-secondary"
-                    data-bs-dismiss="modal"
-                >
-                    Close
-                </button>
-                <button
-                    type="button"
-                    class="btn btn-outline-light"
-                    data-bs-dismiss="modal"
-                    onclick="createRollout()"
-                >
-                    Save changes
-                </button>
-            </div>
+            <form id="rollout-form" class="needs-validation" novalidate>
+                <div class="modal-body">
+                    <div class="form-group mb-3">
+                        <label for="rollout-selected-name">Name</label>
+                        <input
+                            id="rollout-selected-name"
+                            class="form-control"
+                            placeholder="Release 1"
+                        />
+                    </div>
+                    <div class="form-group mb-3">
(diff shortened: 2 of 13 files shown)


goosebit/api/firmware.py
```diff
@@ -25,13 +25,27 @@ def search_filter(search_value):
...
+
+        result = {}
+        for entry in compat:
+            model = entry["model"]
+            revision = entry["revision"]
+            if model not in result:
+                result[model] = []
+            result[model].append(revision)
+
+        compatibility = "<br>".join([f"{model} - {', '.join(revisions)}" for model, revisions in result.items()])
```
b-rowan, line 38: I think this might be better written as `dict[model, set[revision]]`.  This begs the question of restructuring the database to align with that format, namely having `Hardware` contain a `ManyToMany` field which would link to a separate model `Revision`.

goosebit/ui/templates/rollouts.html
```diff
@@ -34,41 +34,67 @@ <h5 class="modal-title">Create Rollout</h5>
...
+                            Feed missing. Use "default" if working with a single
+                            feed.
+                        </div>
+                    </div>
+                    <div class="form-group mb-3">
+                        <label for="rollout-selected-flavor">Flavor</label>
+                        <input
+                            id="rollout-selected-flavor"
+                            class="form-control"
+                            placeholder="debug"
```
b-rowan, line 65: Placeholder seems wrong.

goosebit/ui/templates/rollouts.html
```diff
@@ -34,41 +34,67 @@ <h5 class="modal-title">Create Rollout</h5>
...
+                            class="form-control"
+                            placeholder="Release 1"
+                        />
+                    </div>
+                    <div class="form-group mb-3">
+                        <label for="rollout-selected-feed">Feed</label>
+                        <input
+                            id="rollout-selected-feed"
+                            class="form-control"
+                            placeholder="qa"
```
b-rowan, line 52: Placeholder seems wrong.

goosebit/ui/templates/rollouts.html
```diff
@@ -34,41 +34,67 @@ <h5 class="modal-title">Create Rollout</h5>
...
-                </button>
-            </div>
+            <form id="rollout-form" class="needs-validation" novalidate>
+                <div class="modal-body">
+                    <div class="form-group mb-3">
+                        <label for="rollout-selected-name">Name</label>
+                        <input
+                            id="rollout-selected-name"
+                            class="form-control"
+                            placeholder="Release 1"
```
b-rowan, line 44: Placeholder seems wrong.

pyproject.toml
```diff
@@ -23,6 +23,7 @@ opentelemetry-distro = "^0.47b0"
 opentelemetry-instrumentation-fastapi = "^0.47b0"
 opentelemetry-exporter-prometheus = "^0.47b0"
 aiocache = "^0.12.2"
+httpx = "^0.27.0"
```
b-rowan, line 26: Why was this moved?  Not seeing it imported anywhere in the non-tests files.

b-rowan (changes requested):

goosebit/ui/templates/rollouts.html
tsagadar: "debug" would be a correct example of a flavor. But we decided to remove the "flavor" feature for an initial release -  so this will go away in a future PR.

goosebit/ui/templates/rollouts.html
tsagadar: Again: a proper example for a feed. Idea for a better example - or should we not use any placeholder?

goosebit/ui/templates/rollouts.html
tsagadar: Another example...

pyproject.toml
tsagadar: It is used in `swdesc.py` (it was already here once - and I moved it back to test dependencies as I also did not see where it was used).

b-rowan:
Placeholders make more sense now, I was thinking something more generic for them, but what you have should do fine for now.

goosebit/api/firmware.py
tsagadar: You mean returning the raw data vs. a pre-formatted string? Or simply that result[model] should not be of type list but rather set?

As for the db model: I don't see a big drawback in the current one - so hesitant to invest time there.

b-rowan: I think just not a big fan of the preformatted string.  I would prefer to isolate the UI from the API as much as possible.

tsagadar: Moved the formatting logic back to javascript (had it there initially)

b-rowan (approved):
:+1:

PR #50, opened 2024-08-07
Firmware upload behavior

tsagadar:
Addressing different issues with firmware upload and deletion.

Main commit is 4d56595

### DIFF
--- a/goosebit/models.py
+++ b/goosebit/models.py
@@ -55,7 +55,9 @@ class Tag(Model):
 class Device(Model):
     uuid = fields.CharField(max_length=255, primary_key=True)
     name = fields.CharField(max_length=255, null=True)
-    assigned_firmware = fields.ForeignKeyField("models.Firmware", related_name="assigned_devices", null=True)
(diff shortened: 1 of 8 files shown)


goosebit/models.py
```diff
@@ -136,5 +138,12 @@ def path(self):
...
-    def local(self):
+    def local(self) -> bool:
         return urlparse(self.uri).scheme == "file"
+
+    @property
+    def path_user(self) -> str:
+        if self.local:
+            return self.path.name
+        else:
+            return self.uri
```
b-rowan, line 149: This might be a bit long, maybe want to use the end of the URI path or something?

b-rowan (approved):

goosebit/models.py
tsagadar: That we had before (just the "filename"). For remote firmware, this filename could always be the same - as the difference is in the directory structure. Also helpful to have one place where the full url can be seen IMHO.

Issue #31, opened 2024-07-25
Docker Support


embetrix:
Is a docker (docker compose) support planned ?

embetrix:
okay thanks, have nice holidays

b-rowan:
Do you guys have a specific way we would want to do this?  I think my main concern here is handling the database properly, we have to make sure that alternate DB backends (like Postgres) are supported.

My ideal scenario is that this project just gets a dockerfile, and then end users could make their own docker compose to handle DB and composing with anything else they want (such as grafana/prometheus).

easybe:
The Docker Compose file(s) in this repo would only serve as example. However, the Dockerfile should be used by CI to build and push an image to a registry.

b-rowan:
Technically I think the request from this issue has been completed.   Docker support is done, and Docker compose is WIP as an example.

Docker image is up on docker hub at https://hub.docker.com/r/upstreamdata/goosebit

easybe:
> Technically I think the request from this issue has been completed. Docker support is done, and Docker compose is WIP as an example.
> 
> Docker image is up on docker hub at https://hub.docker.com/r/upstreamdata/goosebit

Correct. If you could push another release I‘d be happy to finish the PR.

b-rowan:
> Correct. If you could push another release I‘d be happy to finish the PR.

Absolutely.  Just a new release whenever to trigger the existing build?

PR #52, opened 2024-08-07
Add Docker (Compose) support

easybe:
A first stab at #31 and #54.

To do:
- [x] Look into CI integration
- [x] Configure linter for YAML (the whole world uses 2 spaces for Docker Compose files)
- [x] Add service user to the image, it is bad practice to run the application as root
- [x] Use `gunicorn`
- [x] Make sure we can run behind a reverse proxy (NGINX)

---

This can be tested with:
```
docker compose up --build
```


Dockerfile
```diff
@@ -0,0 +1,9 @@
+FROM python:3.12.4-alpine
+
+COPY . /tmp/src
+RUN pip install --no-cache-dir /tmp/src && \
```
easybe, line 4: For the image that we upload to the registry I would rather install the package from PyPI here.

pyproject.toml
```diff
@@ -24,6 +24,12 @@ opentelemetry-instrumentation-fastapi = "^0.47b0"
 opentelemetry-exporter-prometheus = "^0.47b0"
 aiocache = "^0.12.2"
 httpx = "^0.27.0"
+dynaconf = "^3.2.6"
+
+asyncpg = { version = "^0.29.0", optional = true }
+
+[tool.poetry.extras]
+postgresql = ["asyncpg"]
```
easybe, line 32: I could not figure out how to reference the `tortoise-orm[asyncpg]` extra here.

b-rowan: I think this is fine for now.

goosebit/settings.py
```diff
@@ -2,23 +2,24 @@
...
 PWD_CXT = PasswordHasher()
 
-with open(BASE_DIR.joinpath("settings.yaml"), "r") as f:
-    config = yaml.safe_load(f.read())
+config = Dynaconf(
+    envvar_prefix="GOOSEBIT",
+    settings_files=[BASE_DIR.joinpath("settings.yaml"), BASE_DIR.joinpath(".secrets.yaml")],
+)
+
+UPDATES_DIR = Path(config.get("artifacts_dir", BASE_DIR.joinpath("updates")))
```
b-rowan, line 22: Might as well rename this from `UPDATES_DIR` to `ARTIFACTS_DIR`.

docker-compose.yml
```diff
@@ -0,0 +1,31 @@
...
+    image: goosebit
+    build:
+      context: .
+      args:
+        GOOSEBIT_VERSION: 0.1.2
+    environment:
+      GOOSEBIT_DB_URI: postgres://postgres:admin@postgres:5432/goosebit
+      GOOSEBIT_USERS: >-
+        [{email="admin@goosebit.local", password="admin", permissions="*"}]
+      GOOSEBIT_ARTIFACTS_DIR: /artifacts
```
b-rowan, line 27: Can you confirm that these settings get imported properly?  I won't be able to test this for a bit yet, so I just want to confirm these actually get loaded properly.

b-rowan (approved):
Looks good overall, just a couple housekeeping/clarification things.

b-rowan:
Regarding CI, I already have an action set up for version releasing, I think we could add a build and push to docker there.

The only concerns I have is we have to figure out how to update the version in the docker compose.

b-rowan:
> Regarding CI, I already have an action set up for version releasing, I think we could add a build and push to docker there.
> 
> The only concerns I have is we have to figure out how to update the version in the docker compose.

Here is a GitHub guide on doing it using actions.  I have added a `DOCKER_USERNAME` and `DOCKER_PASSWORD` value to the secrets for the repo, so those should be usable.  This will push to the `upstreamdata` docker page on docker hub.

docker-compose.yml
easybe: Yes, everything works.

easybe:
> Regarding CI, I already have an action set up for version releasing, I think we could add a build and push to docker there.
> 
> The only concerns I have is we have to figure out how to update the version in the docker compose.

Yes, that is a tricky one. It will be hard to avoid having the version in 2 places. However, the Compose file shall only be used as an example or for development. Build and push I plan to do with GitHub Actions directly.

goosebit/settings.py
easybe: I would then also rename the `updates` directory to `artifacts`. I like the name better, as the files don't necessarily have to be updates. Is that an option for you?

b-rowan: > I would then also rename the `updates` directory to `artifacts`. I like the name better, as the files don't necessarily have to be updates. Is that an option for you?

:100: this is much better.

b-rowan:
>  Configure linter for YAML (the whole world uses 2 spaces for Docker Compose files)

I generally like to keep things consistent across the repo, meaning 4 spaces indentation for all files.  I'm not strictly opposed to this for yaml files, but something we should do some deliberation about.

For reference, this is handled by `prettier`, which does both HTML and YAML files.  I want to keep HTML at 4 spaces, but if we wanted to handle YAML differently we would have to figure out how to in `.prettierrc.yaml`.

b-rowan:
> > Configure linter for YAML (the whole world uses 2 spaces for Docker Compose files)
> 
> I generally like to keep things consistent across the repo, meaning 4 spaces indentation for all files. I'm not strictly opposed to this for yaml files, but something we should do some deliberation about.
> 
> For reference, this is handled by `prettier`, which does both HTML and YAML files. I want to keep HTML at 4 spaces, but if we wanted to handle YAML differently we would have to figure out how to in `.prettierrc.yaml`.

https://stackoverflow.com/questions/56464727/different-prettier-rules-for-different-file-types

easybe:
> > Configure linter for YAML (the whole world uses 2 spaces for Docker Compose files)
> 
> I generally like to keep things consistent across the repo, meaning 4 spaces indentation for all files. I'm not strictly opposed to this for yaml files, but something we should do some deliberation about.
> 
> For reference, this is handled by `prettier`, which does both HTML and YAML files. I want to keep HTML at 4 spaces, but if we wanted to handle YAML differently we would have to figure out how to in `.prettierrc.yaml`.

I get your point. I generally also use 4 spaces. However, as pretty much all YAML out there uses 2 spaces, my eyes got used to it and it makes copying & pasting from the internet easier.

b-rowan:
> I get your point. I generally also use 4 spaces. However, as pretty much all YAML out there is indented with 2 spaces, my eyes got used to it, but more importantly, it makes copying & pasting from the internet easier.

Ok, let's move forward with 2 space yaml then.  I think that link I provided should be enough info to configure prettier, though it will need to be converted into yaml.

easybe:
> > I get your point. I generally also use 4 spaces. However, as pretty much all YAML out there is indented with 2 spaces, my eyes got used to it, but more importantly, it makes copying & pasting from the internet easier.
> 
> Ok, let's move forward with 2 space yaml then. I think that link I provided should be enough info to configure prettier, though it will need to be converted into yaml.

#59

easybe:
Concerning CI, what I got so far seems to work: https://github.com/UpstreamDataInc/goosebit/actions/runs/10322906355

To test the actual push to the registry is a little bit of a challenge though, as the workflow is only meant to be run on release.

I will come up with a strategy and start getting things ready for integration.

b-rowan:
If you only want to run on release, I use this file for release to push to PyPi- https://github.com/UpstreamDataInc/goosebit/blob/master/.github/workflows/python-publish.yml

easybe:
@b-rowan I would assume that the [upstreamdata/goosebit](https://hub.docker.com/r/upstreamdata/goosebit) repository needs to be manually created on Docker Hub.

b-rowan:
Ok, let me try doing that, one moment.

.github/workflows/docker.yml
```diff
@@ -0,0 +1,63 @@
+name: Docker
+
+# on:
+#   workflow_run:
+#     workflows: ["PyPI"]
+#     types:
+#       - completed
```
easybe, line 7: > If you only want to run on release, I use this file for release to push to PyPi- https://github.com/UpstreamDataInc/goosebit/blob/master/.github/workflows/python-publish.yml

Yes, I know. For that, I prepared this.

b-rowan:
Ok, repo is up on docker hub :+1:

easybe:
I will split this into multiple PRs. First one: 60

b-rowan:
> Use gunicorn

Here's a reference from the FastAPI docs on this, basically just run with uvicorn workers.

https://fastapi.tiangolo.com/deployment/server-workers/#gunicorn-with-uvicorn-workers

easybe:
> > Use gunicorn
> 
> Here's a reference from the FastAPI docs on this, basically just run with uvicorn workers.
> 
> https://fastapi.tiangolo.com/deployment/server-workers/#gunicorn-with-uvicorn-workers

Yes, that is what am working. Unfortunately, I am still struggling with getting the logger to print the format I want (and lack of time as I am also working on our internal deployment). But, I am on it. My WIP branch can be found here: https://github.com/husqvarnagroup/goosebit/tree/gardena/eb/docker_compose

b-rowan:
> > > Use gunicorn
> 
> > 
> 
> > Here's a reference from the FastAPI docs on this, basically just run with uvicorn workers.
> 
> > 
> 
> > https://fastapi.tiangolo.com/deployment/server-workers/#gunicorn-with-uvicorn-workers
> 
> 
> 
> Yes, that is what am working on. Unfortunately, I am still struggling with getting the logger to print the format I want (and lack of time as I am also working on our internal deployment). But, I am on it. My WIP branch can be found here: https://github.com/husqvarnagroup/goosebit/tree/gardena/eb/docker_compose

All good, just wanted to help out, as I figured y'all have been busy.

easybe:
Closing in favor of #88 and #109

PR #57, opened 2024-08-08
docs: auto deploy to gh-pages.

### LINKED ISSUE #56: Search engine optimization (SEO)
When I google “goosebit”, our fork shows up and not the main repo, not even when I search for “goosebit github”. Any idea why that could be?

Also, https://upstreamdata.com/open-source-projects/goosebit/ shows up at the top, which looks very unfinished…

b-rowan:
Closes #56.

### DIFF
--- a/.github/workflows/gh-pages.yml
+++ b/.github/workflows/gh-pages.yml
@@ -1,0 +1,36 @@
+name: GitHub Pages
+
+on:
+  push:
+    branches:
+      - "master"
+    paths:
+      - ".github/workflows/update-docs.ya?ml"
+      - "mkdocs.ya?ml"
+      - "poetry.lock"
+      - "pyproject.toml"
+      - "docs/**"
+
+jobs:
+  publish:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      - name: Install Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: 3.12
+
+      - name: Install Poetry
+        run: pip install poetry
+
+      - name: Install Dependencies
+        run: poetry install --no-root --with docs
+
+      - name: Deploy website
+        run: poetry run mkdocs gh-deploy


.github/workflows/update-docs.yml
```diff
@@ -0,0 +1,36 @@
+name: PyPI
```
easybe, line 1: I would just call the file `docs.yml`, without a verb.

b-rowan: Renamed to `gh-pages.yml`, which makes more sense since im already auto-deploying to readthedocs (https://goosebit.rtfd.io)

.github/workflows/gh-pages.yml
easybe, line 1: Any particular reason you don’t use `git commit —amend` or `git rebase —interactive`? Such commits are pure noise.

b-rowan: Mostly because I don't know how to use git :laughing: 

Also, I did that commit via web interface because I didn't want to have to stash changes on the other branch I'm working on right now.   It all gets squashed in the end anyway right?

easybe: No need to stash, if you commit often and amend/squash (with interactive rebase) after. I recommended looking into that, having a clean Git history can be very beneficial. Some more info on that you can find e.g [here](https://cbea.ms/git-commit/).

Weather commits are preserved or not depends on the strategy you use for merging PRs. I personally prefer „rebase only“. However, that only really works with clean commits. „Squash and merge“ I wouldn’t use, having multiple (clean) commits can drastically improve the readability of the history.

b-rowan: > No need to stash, if you commit often and amend/squash (with interactive rebase) after. I recommended looking into that, having a clean Git history can be very beneficial. Some more info on that you can find e.g [here](https://cbea.ms/git-commit/).

This seems like a useful resource, thank you.  Right now I've been too lazy to really try to use CLI git, I just use the pycharm built ins (which work quite well thankfully), but maybe ill have to learn.

> Whether commits are preserved or not depends on the strategy you use for merging PRs. I personally prefer „rebase only“. However, that only really works with clean commits. „Squash and merge“ I wouldn’t use, having multiple (clean) commits can drastically improve the readability of the history.

Works for me, my only thought was it might be nice when merging to be able to have a commit that is sortof "atomic" to that PR, in that it describes exactly what the PR aimed to do, without too much verbosity.

b-rowan: I think I fixed it?

easybe: > > No need to stash, if you commit often and amend/squash (with interactive rebase) after. I recommended looking into that, having a clean Git history can be very beneficial. Some more info on that you can find e.g [here](https://cbea.ms/git-commit/).
> 
> This seems like a useful resource, thank you. Right now I've been too lazy to really try to use CLI git, I just use the pycharm built ins (which work quite well thankfully), but maybe ill have to learn.
> 
> > Whether commits are preserved or not depends on the strategy you use for merging PRs. I personally prefer „rebase only“. However, that only really works with clean commits. „Squash and merge“ I wouldn’t use, having multiple (clean) commits can drastically improve the readability of the history.
> 
> Works for me, my only thought was it might be nice when merging to be able to have a commit that is sortof "atomic" to that PR, in that it describes exactly what the PR aimed to do, without too much verbosity.

I tend to sneak unrelated stuff into PRs as I think in commits, which IMHO are preferably atomic but more importantly reviewable. But, that‘s a matter of taste I guess.

.github/workflows/gh-pages.yml
```diff
@@ -0,0 +1,36 @@
+name: PyPI
```
easybe, line 1: This name should be different, no?

easybe (changes requested):

.github/workflows/gh-pages.yml
```diff
@@ -0,0 +1,36 @@
...
+            contents: write
+
+        steps:
+            - name: Checkout repository
+              uses: actions/checkout@v4
+
+            - name: Install Python
+              uses: actions/setup-python@v5
+              with:
+                  python-version: ${{ env.PYTHON_VERSION }}
```
easybe, line 27: Where is `PYTHON_VERSION` set?

b-rowan: Using static `3.12` now.

.github/workflows/gh-pages.yml
```diff
@@ -0,0 +1,36 @@
+name: GitHub Pages
+
+on:
+    push:
```
easybe, line 4: You should rebase (`git rebase origin/master && git push --force-with-lease`) and run `pre-commit run --all-files`.

.github/workflows/gh-pages.yml
```diff
@@ -0,0 +1,36 @@
...
+            contents: write
+
+        steps:
+            - name: Checkout repository
+              uses: actions/checkout@v4
+
+            - name: Install Python
+              uses: actions/setup-python@v5
+              with:
+                  python-version: "3.12"
```
easybe, line 27: I don't think quotes are needed here.

easybe (changes requested):

.github/workflows/gh-pages.yml
easybe: OK. But where was it defined first? :confused:

b-rowan: ...

Probably wherever I copy pasted it from 😅🤣

easybe (approved):

PR #64, opened 2024-08-12
Add support for configuration through env vars

easybe:
In addition to the key-value pairs in settings.yaml, the application can now be configured through corresponding `GOOSEBIT_*` environment variables. E.g. to use PostgreSQL, set:

    GOOSEBIT_DB_URI=postgres://db_user:db_pw@postgres:5432/goosebit

The users are configured as follows:

    GOOSEBIT_USERS='[{email="foo@b.ar", password="42", permissions="*"}]'

The environment variables have precedence over the configuration file.

For more information, read: https://www.dynaconf.com/envvars/

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -167,3 +167,6 @@ cython_debug/
 /db.sqlite3*
 /migrations
 /poetry.lock
+
+# dynaconf secret files
+.secrets.*
--- a/goosebit/settings.py
+++ b/goosebit/settings.py
@@ -2,8 +2,8 @@ import secrets
 from dataclasses import dataclass
 from pathlib import Path
 
-import yaml
 from argon2 import PasswordHasher
+from dynaconf import Dynaconf
 from joserfc.jwk import OctKey
 
 from goosebit.permissions import Permissions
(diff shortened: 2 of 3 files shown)


b-rowan (approved):
Looks good.  Had to check docs to make sure it would parse them in as lower case variables, but it seems like it does, so should just be drop-in.

b-rowan:
@easybe if you want to merge this go for it, I think this is in a good enough spot for now, and dropping in pydantic later shouldn't be a breaking change, just a matter of when we want to get rid of that extra dependency.

easybe:
Actually, switching to pydantic might be a breaking change as probably

    GOOSEBIT_USERS='[{email="foo@b.ar", password="42", permissions="*"}]'

will no longer work. So, we should probably use it now if we know that it is the path we want to take.

b-rowan:
Scrolling down a bit, it seems like should still work, we just have an item typed as `list[User]` or whatever, and it should parse from that.

Copied the example cause there wasn't an anchor nearby...

```

import os

from pydantic import BaseModel, ValidationError

from pydantic_settings import BaseSettings


class RedisSettings(BaseModel):
    host: str
    port: int


class Settings(BaseSettings, case_sensitive=True):
    redis: RedisSettings


os.environ['redis'] = '{"host": "localhost", "port": 6379}'
print(Settings().model_dump())
#> {'redis': {'host': 'localhost', 'port': 6379}}
os.environ['redis'] = '{"HOST": "localhost", "port": 6379}'  
try:
    Settings()
except ValidationError as e:
    print(e)
    """
    1 validation error for Settings
    redis.host
      Field required [type=missing, input_value={'HOST': 'localhost', 'port': 6379}, input_type=dict]
        For further information visit https://errors.pydantic.dev/2/v/missing
    """
```

b-rowan:
I think this shows it accurately enough - https://docs.pydantic.dev/latest/concepts/pydantic_settings/#parsing-environment-variable-values

easybe:
My example uses the TOML format, yours JSON.

easybe:
Rendered obsolete by #71

PR #66, opened 2024-08-13
Move download to a separate router

b-rowan:
Moves the download endpoints to a separate endpoint (`/download/{file}`) to prevent the weird multi-registration in API.


b-rowan:
Superseded by #58.

PR #71, opened 2024-08-15
Use pydantic for settings.

b-rowan:
Closes #64.


goosebit/settings/__init__.py
```diff
@@ -0,0 +1,18 @@
+from .const import BASE_DIR, DB_MIGRATIONS_LOC, PWD_CXT, SECRET  # noqa: F401
+from .schema import GooseBitSettings
+
+config = GooseBitSettings()
+print(config)
```
easybe, line 5: These print statements should be removed, no?

goosebit/settings/__init__.py
```diff
@@ -0,0 +1,18 @@
...
+POLL_TIME = config.poll_time_default
+POLL_TIME_UPDATING = config.poll_time_updating
+POLL_TIME_REGISTRATION = config.poll_time_registration
+DB_URI = config.db_uri
+USERS = {u.username: u for u in config.users}
+LOGGING = config.logging
+
+PROMETHEUS_PORT = config.metrics.prometheus.port
+
+print(USERS)
```
easybe, line 18: rm

goosebit/settings/__init__.py
```diff
@@ -0,0 +1,18 @@
+from .const import BASE_DIR, DB_MIGRATIONS_LOC, PWD_CXT, SECRET  # noqa: F401
+from .schema import GooseBitSettings
+
+config = GooseBitSettings()
+print(config)
+
+ARTIFACTS_DIR = config.artifacts_dir
```
easybe, line 7: Does it make sense to keep these constants? Why not just use `config.artifacts_dir` directly in the code?

easybe (commented):
Thanks!

goosebit/settings/__init__.py
b-rowan: Ha yeah, might need a pre-commit linter for `get rid of random print statements`...

goosebit/settings/__init__.py
b-rowan: Totally fine with that.  Let me update to that then.

goosebit/settings/schema.py
```diff
@@ -0,0 +1,74 @@
...
+class PrometheusSettings(BaseModel):
+    port: int = 9090
+
+
+class MetricsSettings(BaseModel):
+    prometheus: PrometheusSettings = PrometheusSettings()
+
+
+class GooseBitSettings(BaseSettings):
+    model_config = SettingsConfigDict(env_prefix="goosebit_")
```
easybe, line 42:
```suggestion
    model_config = SettingsConfigDict(env_prefix="GOOSEBIT_")
```

Even though this shouldn't make a difference, it might for the reader. Envronment variables should be ALL_CAPS.

b-rowan: Environment variables when parsed with pydantic are case-insensitive, we can change this but it will definitely not make any difference.

goosebit/settings/schema.py
```diff
@@ -39,7 +39,7 @@ class MetricsSettings(BaseModel):
 
 
 class GooseBitSettings(BaseSettings):
-    model_config = SettingsConfigDict(env_prefix="goosebit_")
```
easybe: Please squash such commits created by the UI (locally) .

easybe: This easily be done as follows:

1. `git rebase -i origin/master`
2. replace `pick` with `f` at the appropriate line.
3. CLose the editor.

goosebit/settings/schema.py
easybe: As I said, it is only for the readers of the code.

goosebit/settings/schema.py
```diff
@@ -31,6 +31,7 @@ def get_json_permissions(self):
 
 
 class PrometheusSettings(BaseModel):
+    enable: bool = True
```
easybe, line 34: IMHO should be off by default as not everyone has Prometheus.

easybe (commented):
LGTM, will test tomorrow

easybe (approved):

PR #77, opened 2024-08-19
Move metrics to `/api` prefix and handle telemetry in API

### LINKED ISSUE #72: Telemetry prevents running multiple instances
https://github.com/UpstreamDataInc/goosebit/blob/c787cb4a30ab1e64daeb50410fed11ffccdd9750/goosebit/telemetry/prometheus.py#L10

This does not work when running multiple instances e.g. with gunicorn workers. We need to keep concurrency in mind.

b-rowan:
Moves telemetry to the `/api` router, and moves api endpoints to `/api/v1`.  Closes #72.


main.py
```diff
@@ -7,7 +7,7 @@
 
 logging.config.dictConfig(config.logging)
 
-uvicorn_args = {"port": 80}
+uvicorn_args = {"port": 8080}
```
tsagadar, line 10: Intentional change?

tsagadar (changes requested):

tsagadar (approved):

PR #78, opened 2024-08-19
Use pydantic models for DDI API

b-rowan:
Adds pydantic schemas for `configData` and `feedback` which mostly match the DDI API docs (aside from `id`, see `TODO`).

### DIFF
--- a/goosebit/updater/controller/v1/routes.py
+++ b/goosebit/updater/controller/v1/routes.py
@@ -1,4 +1,3 @@
-import json
 import logging
 
 from fastapi import APIRouter, Depends, HTTPException
@@ -10,18 +9,20 @@ from goosebit.settings import config
 from goosebit.updater.manager import HandlingType, UpdateManager, get_update_manager
 from goosebit.updates import generate_chunk
 
+from .schema import (
+    ConfigDataSchema,
+    FeedbackSchema,
+    FeedbackStatusExecutionState,
+    FeedbackStatusResultFinished,
+)
+
 logger = logging.getLogger("DDI API")
 
 router = APIRouter(prefix="/v1")
 
 
 @router.get("/{dev_id}")
-async def polling(
-    request: Request,
-    tenant: str,
-    dev_id: str,
-    updater: UpdateManager = Depends(get_update_manager),
-):
+async def polling(request: Request, tenant: str, dev_id: str, updater: UpdateManager = Depends(get_update_manager)):
     links = {}
 
     sleep = updater.poll_time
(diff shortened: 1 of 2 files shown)


goosebit/updater/controller/v1/routes.py
```diff
@@ -106,77 +98,66 @@ async def deployment_base(
...
+
+    elif data.status.execution in [
+        FeedbackStatusExecutionState.CLOSED,
+        FeedbackStatusExecutionState.CANCELED,
+        FeedbackStatusExecutionState.REJECTED,
+    ]:
+        await updater.update_force_update(False)
+        await updater.update_log_complete(True)
+
+        # TODO: check this, not in DDI schema
```
b-rowan, line 115: This is not mentioned anywhere in the hawkbit DDI docs, but I added this just in case.

tsagadar: DDI just talks about an abstract action_id. gooseBit could create for each update a new record and use its id to track the update. To prevent huge tables (and slow processing) I decided to use the firmware_id as the action_id. This allows to precisely track which firmware got installed.

tsagadar (changes requested):

main.py
```diff
@@ -7,7 +7,7 @@
 
 logging.config.dictConfig(config.logging)
 
-uvicorn_args = {"port": 80}
+uvicorn_args = {"port": 8080}
```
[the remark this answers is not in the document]
b-rowan: Good catch, my desktop won't let me bind to port 80 (at least not without a fight). I'll remove that.

goosebit/updater/controller/v1/routes.py
```diff
@@ -106,77 +98,66 @@ async def deployment_base(
...
-
-    except KeyError as e:
-        logging.warning(f"Processing deployment feedback failed, error={e}, device={updater.dev_id}")
+    if data.status.execution == FeedbackStatusExecutionState.PROCEEDING:
+        await updater.update_device_state(UpdateStateEnum.RUNNING)
+        logger.debug(f"Installation in progress, device={updater.dev_id}")
+
+    elif data.status.execution in [
+        FeedbackStatusExecutionState.CLOSED,
+        FeedbackStatusExecutionState.CANCELED,
```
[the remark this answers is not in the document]
b-rowan: Makes sense.  I'll add logging for any other unhandled states.

goosebit/updater/controller/v1/routes.py
b-rowan: Oh, the ID isn't getting passed through the API request, it's getting passed as a variable in the URL?  That makes a lot more sense, I forgot that change was made.  I'll remove this and make sure that's referenced properly.

tsagadar (approved):

PR #79, opened 2024-08-23
Add proper API token usage

### LINKED ISSUE #70: API access via token

### LINKED ISSUE #73: Session does not seem to be shared between gunicorn workers
This can be seen when running e.g. with:
```
gunicorn --workers 8 --worker-class uvicorn.workers.UvicornWorker --bind 0.0.0.0:8080 goosebit:app
```

I assume the session data needs to be stored in the DB.

b-rowan:
Adds API tokens in `authorization` header, and allows setting `secret_key` to allows the tokens to persist across application restarts.

Closes #70 
Closes #73


goosebit/__init__.py
```diff
@@ -28,7 +24,24 @@ async def lifespan(_: FastAPI):
...
+app = FastAPI(
+    title="gooseBit",
+    summary="A simplistic, opinionated remote update server implementing hawkBit™'s DDI API.",
+    version=importlib.metadata.version("goosebit"),
+    lifespan=lifespan,
+    license_info={
+        "name": "Apache 2.0",
+        "identifier": "Apache-2.0",
+    },
+    swagger_ui_parameters={"operationsSorter": "alpha"},
```
easybe, line 36: Currently, we have Swagger UI (/docs) and ReDoc (/redoc). I think ReDoc looks quite nice but, doesn't seem to show the schemas. Anyway, I think we should decide on one and disable the other.

easybe (approved):
LGTM

Thank you very much!

easybe:
How does one retrieve the token and pass it in an HTTP header? I would expect to find this information in the commit message body. You should really consider reading https://cbea.ms/git-commit/ :pray:

easybe:
OK, got it. One gets the token by POSTing to `/login`. That works for me. Still, would appreciate better commit messages.

b-rowan:
> OK, got it. One gets the token by POSTing to `/login`. That works for me.
> 
> 
> 
> Still, would appreciate better commit messages.

Also, along with the token query, you can also now set the secret in the settings file.  This means that once you query the token it is static for any given user.

Regarding commit messages, you're asking for more context in the commit body?

goosebit/__init__.py
b-rowan: My preference is swagger, mainly because of the schemas as mentioned.  I will try to put forward another PR soon that will customize the swagger UI a bit further, including the favicon, and I can remove redoc there.

Issue #74, opened 2024-08-16
Firmware vs Software


easybe:
Throughout gooseBit the term firmware is used, whereas the Linux counterpart, SWUpdate uses the term software. Another "problem" I see is that in the Linux world firmware is normally what is written/flashed to peripherals or chips on the board.

Would it be an option to use "software" instead?

b-rowan:
> Throughout gooseBit the term firmware is used, whereas the Linux counterpart, SWUpdate uses the term software. Another "problem" I see is that in the Linux world firmware is normally what is written/flashed to peripherals or chips on the board.
> 
> Would it be an option to use "software" instead?

No opposition, I'm just very used to anything flashed onto an embedded device being called "firmware", but that might be avery niche viewpoint.

easybe:
OK. So maybe achieve some stability first? Preferably have some tests. Then do a big find & replace?

b-rowan:
> OK. So maybe achieve some stability first? Preferably have some tests. Then do a big find & replace?

Sounds good.  Are we going to aim for 100% test coverage of the project on this note?

tsagadar:
100% test coverage IMHO rarely makes sense. I'd focus on the DDI parts to have this covered as much as possible - here things could go wrong and in the worst case brick devices (or not detecting, that a new software bricks them).

On the UI side it is annoying to have bugs, but not much harm done.

PR #84, opened 2024-08-24
Rename firmware to software / remove tenant

tsagadar:
Addressing two things that impact our endpoints
1. Rename firmware to software (#74)
2. Remove tenant concept (replace DEFAULT by ddi in endpoints)


goosebit/updater/routes.py
```diff
@@ -1,29 +1,21 @@
...
 async def log_last_connection(request: Request, dev_id: str):
     host = request.client.host
     updater = await get_update_manager(dev_id)
     await updater.update_last_connection(round(time.time()), host)
 
 
 router = APIRouter(
-    prefix="/{tenant}",
-    dependencies=[Depends(verify_tenant), Depends(log_last_connection)],
+    prefix="/ddi",
```
b-rowan, line 17: I suppose in the future we may want to have this just be a variable, then we can introduce `tenant` as another firmware selection layer?  Not sure if it will be better than what we have with device IDs and rollouts, but maybe it will allow automating the initial selection of those variables?

b-rowan (approved):
Just a couple things for discussion.

goosebit/updater/routes.py
easybe: IMHO it would be fine to just change the default to „ddi“. The advantage of leaving (configurable) tenant support would be that devices currently connected to hawkBit would not have to be changed for gooseBit.

easybe:
@tsagadar thanks for the birthday gift 😂

README.md
```diff
@@ -21,9 +21,9 @@ The default login credentials for testing are `admin@goosebit.local`, `admin`.
 
 ## Current Feature Set
 
-### Firmware repository
+### Software repository
 
-Uploading firmware images through frontend. All files should follow the format `{model}_{revision}_{version}`, where
+Uploading software images through frontend. All files should follow the format `{model}_{revision}_{version}`, where
```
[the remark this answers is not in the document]
tsagadar: Will update README.md in another PR.

docs/index.md
```diff
@@ -21,9 +21,9 @@ The default login credentials for testing are `admin@goosebit.local`, `admin`.
 
 ## Current Feature Set
 
-### Firmware repository
+### Software repository
 
-Uploading firmware images through frontend. All files should follow the format `{model}_{revision}_{version}`, where
+Uploading software images through frontend. All files should follow the format `{model}_{revision}_{version}`, where
```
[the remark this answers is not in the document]
tsagadar: Currently we struggle to keep the README.md up-to-date. Not sure if we already need this detailed documentation. But will mirror the README.md updates here as well.

b-rowan:
LGTM, all good now.

goosebit/updater/routes.py
tsagadar: Overlooked this comment before merging - sorry. Still would strive for simplicity by hard-coding it.

PR #85, opened 2024-08-26
Updates on README.md

tsagadar:
Updating all the outdated information to catch up with the recent development.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,55 +6,68 @@
 
 A simplistic, opinionated remote update server implementing hawkBit™'s [DDI API](https://eclipse.dev/hawkbit/apis/ddi_api/).
 
-## Setup
+## Quick Start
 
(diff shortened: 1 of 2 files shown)


b-rowan (approved):

b-rowan:
I think it might be time for `0.2.0` after this gets merged, thoughts?

tsagadar:
As for the releasing: I have not done tests with real devices for a bit. Waiting for the docker installation from @easybe before doing that work. Not sure how close we are there though? If it still takes a bit, I can do some tests locally tonight (CET).

easybe:
> As for the releasing: I have not done tests with real devices for a bit. Waiting for the docker installation from @easybe before doing that work. Not sure how close we are there though? If it still takes a bit, I can do some tests locally tonight (CET).

Our internal deployment will be updated today.

PR #86, opened 2024-08-26
Improve permission handling

b-rowan:
Improve permissions handling with string parsing.

Rules are as follows - 
- `*` - Matches all at all levels.
- `!` - Prefix, inverts the permission, and takes priority.
- `foo` - The name of a permission, just `foo` on its own grants all sub permissions for that value.
- `foo.*` - Same as `foo`.
- `foo.bar` - Grants `bar` permissions for `foo` endpoint.

For example, if an endpoint requires `foo.bar.baz`, valid user permissions are 
- `*`
- `foo`
- `foo.*`
- `foo.bar`
- `foo.bar.*`
- `foo.bar.baz`

User permissions that will always be invalid are the inverted versions of this, so 
- `!*`
- `!foo`
- `!foo.*`
- `!foo.bar`
- `!foo.bar.*`
- `!foo.bar.baz`


b-rowan:
On second thought, leaving as a draft until #85 gets merged, so I can add docs on this.

main.py
```diff
@@ -7,7 +7,7 @@
 
 logging.config.dictConfig(config.logging)
 
-uvicorn_args = {"port": 80}
+uvicorn_args = {"port": 8080}
```
tsagadar, line 10: Our good old friend again ;-)

goosebit/api/v1/devices/device/routes.py
```diff
@@ -5,23 +5,22 @@
...
-from goosebit.permissions import Permissions
 from goosebit.updater.manager import UpdateManager, get_update_manager
 
 router = APIRouter(prefix="/{dev_id}")
 
 
 @router.get(
     "",
-    dependencies=[Security(validate_user_permissions, scopes=[Permissions.HOME.READ])],
+    dependencies=[Security(validate_user_permissions, scopes=["home.read"])],
```
tsagadar, line 15: Bit sad to switch from a "type-safe" approach to free form strings that can have typos. But ok for me.

goosebit/auth/__init__.py
```diff
@@ -117,15 +117,19 @@ def validate_user_permissions(
 
 
 def compare_permissions(scopes: list[str] | None, permissions: set[str]) -> bool:
+    deny_permissions = [p.replace("!", "") for p in permissions if p.startswith("!")]
```
tsagadar, line 120: Could be a good area to write a few unit tests (like https://github.com/UpstreamDataInc/goosebit/blob/master/tests/updates/test_swdesc.py). This could encode the description that comes with this PR and guarantee that it also works.

main.py
easybe: I suggest we revert commit c395cbe. Using port 80 was never a good idea in the first place.

goosebit/auth/__init__.py
easybe, line 120:
```suggestion
    deny_permissions = list(map(lambda x: x.lstrip(“!”), permissions))
```

main.py
tsagadar: So basically having macOS users suffer and not Linux users ;-) Ideally we would revert and have a mechanism to overwrite this setting with an .env file.

easybe: I wouldn't call your use case suffering.

Anyway, with the planned Docker Compose support, you will get to easily run a usable server on your machine.

Alternatively, you can also do the following:
```
python3 -m venv venv
. venv/bin/activate
python -m pip install -e .

uvicorn --port 8080 --reload goosebit:app
```

b-rowan: > So basically having macOS users suffer and not Linux users ;-) Ideally we would revert and have a mechanism to overwrite this setting with an .env file.

I'm using Arch, and this is still a problem by default...

goosebit/api/v1/devices/device/routes.py
b-rowan: Yeah it's not my favorite either, but I think this makes more sense for allowing plugins.

I could try to make this more type safe, maybe by having the nav register permissions, then checking if all user permissions match a valid pattern, but that could be a problem with plugins.

goosebit/auth/__init__.py
b-rowan: Good call, I'll see what I can do here.

main.py
b-rowan: How about port `60053`?  Best port approximation for `GOOSE` ;)

easybe: It is indeed a funny idea. However, IMHO there is nothing wrong with 8080, it is a common choice for the use case. In my experience, ports in the 60k range are rather uncommon.

b-rowan: > It is indeed a funny idea. However, IMHO there is nothing wrong with 8080, it is a common choice for the use case. In my experience, ports in the 60k range are rather uncommon.

Fair enough.  Figure we might as well commit to the bit though.

My biggest concern is 8080 is usually pretty common for testing purposes, but not a big deal in production.

How about we just add it to the config file (no idea why its not in there), and then it can be passed there or as an ENV variable.

b-rowan: Handled with #92

goosebit/auth/__init__.py
b-rowan: Attempted this, let me know how they are.  Not very good at writing tests, but I think these should adequately test the functionality.

tests/auth/test_permissions.py
```diff
@@ -0,0 +1,51 @@
+from goosebit.auth import compare_permission, compare_permissions
```
tsagadar, line 1: Great to have tests! They will help to ensure that this critical function does not get broken by refactoring. Also a good docu to understand what is possible.

Reading the tests I felt that `compare` likely is the wrong term - compare functions usually return "smaller", "equal", "bigger" as int. Should it rather be `check`?

Also: most tests focus on `compare_permission`. This seems a helper function only used internally to the module. So I'd rename it to `_compare_permission` (well: at the end: `_check_permission`) and change the tests to focus on the public function that is used externally to the module.

tsagadar (approved):

PR #90, opened 2024-08-27
Fix software deletion from UI

### LINKED ISSUE #89: Software enty (for external artifact) cannot be deleted
When trying to delete a software entry for an external URL I get the following error message:

Screenshot from 2024-08-19 14-25-17

Int the web browser's console I get:
```
util.js:102     DELETE https://goosebit.foo.net/api/v1/software 422
delete_request @ util.js:102
deleteSoftware @ software.js:269
action @ software.js:151
a @ datatables.min.js:30
r @ datatables.min.js:30
(anonymous) @ datatables.min.js:30
dispatch @ datatables.min.js:14
v.handle @ datatables.min.js:14
software.js:272 Deleting software list failed: Error: POST /api/v1/software failed for [1]
    at delete_request (util.js:119:15)
    at async deleteSoftware (software.js:269:9)
    at async U.action (software.js:151:17)
``` 

The associated rollout entry was deleted successfully.

tsagadar:
UI sends array of IDs as the payload, but backend expected object with files property. Files is not a good naming, so adjusted backend.

UI however should always call the BFF, so some more work required. Not creating a test case for this issue before this bigger refactoring.

Fixes #89

### DIFF
--- a/goosebit/api/v1/software/requests.py
+++ b/goosebit/api/v1/software/requests.py
@@ -1,5 +1,5 @@
-from pydantic import BaseModel
+from pydantic import RootModel
 
 
-class SoftwareDeleteRequest(BaseModel):
-    files: list[int]
+class SoftwareDeleteRequest(RootModel[list[int]]):
+    pass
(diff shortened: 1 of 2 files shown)


b-rowan (approved):
Not sure we want to go the `RootModel` route, we will have to make that decision later.  For now this should be fine.

PR #94, opened 2024-08-27
Use only BFF from UI

### LINKED ISSUE #81: API for CI integration
I would like to be able to do the following through the API:

**Upload firmware**

I would expect to be able to POST to `/api/v1/firmware` and either provide a path or an URL to a file.

I guess something like this could work:

```python
from fastapi import File, UploadFile, Form
(...)
async def firmware_post(
    file: Optional[UploadFile] = File(None),
    url: Optional[str] = Form(None)
):
    if file:
        # Process the uploaded file
    elif url:
        # Process the firmware URL
    else:
        # Return an error
```

**Query firmware**

Get the ID of a firmware entry by providing the URL/filename:
```
GET /api/v1/firmware?url=https://...
```

### LINKED ISSUE #91: Ensure UI only calls BFF endpoints
As discussed here https://github.com/UpstreamDataInc/goosebit/pull/58#discussion_r1717919294, we wanted the UI to only call `/ui/bff` - never `/api/v1`. Somehow I missed to this is not yet implemented, so creating a ticket for it here.

Personally, I would only expose `/api/v1` endpoints that are currently required to integrate gooseBit the surrounding infrastructure (like uploading software from a CI). This because `/api/v1` should be a stable interface. The wider it is, the more likely we will break it.

In `/ui/bff` we have more freedom to change things around, so we also don't need the optimal structure right from the start but will find it over time.

b-rowan:
Refactors UI and BFF to not require any use of API from the UI.  BFF is basically a copy of API for now (except for the `GET` methods).

Also some misc JS fixes, and an added router for uploading firmware.  Uploading firmware is now combined into 1 endpoint in both the API and BFF.

Closes #81, Closes #91


goosebit/api/v1/software/requests.py
```diff
@@ -1,5 +1,5 @@
-from pydantic import RootModel
+from pydantic import BaseModel
 
 
-class SoftwareDeleteRequest(RootModel[list[int]]):
-    pass
+class SoftwareDeleteRequest(BaseModel):
+    files: list[int]
```
tsagadar, line 5: Naming is misleading: it is not really a list of files, but rather a list of software ids.

goosebit/api/v1/software/routes.py
```diff
@@ -25,7 +25,7 @@ async def software_get(_: Request) -> SoftwareResponse:
 )
 async def software_delete(_: Request, config: SoftwareDeleteRequest) -> StatusResponse:
     success = False
-    for f_id in config.root:
+    for f_id in config.files:
```
tsagadar, line 31: `f_id` then also would be the wrong variable naming.

goosebit/ui/bff/software/routes.py
```diff
@@ -23,3 +33,72 @@ def search_filter(search_value):
     total_records = await Software.all().count()
 
     return await BFFSoftwareResponse.convert(request, query, search_filter, total_records)
+
+
+@router.delete(
+    "",
+    dependencies=[Security(validate_user_permissions, scopes=["software.delete"])],
+)
+async def software_delete(_: Request, files: SoftwareDeleteRequest) -> StatusResponse:
```
tsagadar, line 40: Code being duplicated. As long as API stays the same It can simply be forwarded?

goosebit/ui/bff/devices/routes.py
```diff
@@ -31,3 +35,38 @@ def search_filter(search_value):
     total_records = await Device.all().count()
 
     return await BFFDeviceResponse.convert(request, query, search_filter, total_records)
+
+
+@router.patch(
+    "",
+    dependencies=[Security(validate_user_permissions, scopes=["device.write"])],
+)
+async def devices_patch(_: Request, config: DevicesPatchRequest) -> StatusResponse:
```
tsagadar, line 43: Like in other places: code duplication. Here I don't think we need to expose the patch functionality as external API (yet) - so best to delete it from out official integration API until such a request would come up.

tsagadar (changes requested):

goosebit/ui/bff/software/routes.py
b-rowan: My only thought here was if we intend to change the API at any point before 1.0.0 this will need to be changed anyway, and I think just leaving the code duplicated and separate might save some mistakes later.  I'm all for just calling one endpoint from the other, but in which direction?  Probably calling BFF from API makes the most sense, if API changes we just change it there, but if BFF changes it's a breaking API change.

tsagadar: I'd done it the other way around: as we should keep the v1 API stable (as external scripts will start using it), I'd added the code there. As long as BFF is happy with that code it can re-use it - if it wants something different, then it has to implement it on its own. 

Breaking changes in BFF are not problematic, as server and client are under our direct control. Breaking changes in the API should not happen, it requires to introduce a v2.

goosebit/ui/bff/devices/routes.py
b-rowan: Removed from API, but since it was used in tests I had to have the tests use the BFF API.

tests/updater/controller/v1/test_routes.py
```diff
@@ -8,7 +8,7 @@
 
 async def _api_device_update(async_client, device, update_attribute, update_value):
     response = await async_client.patch(
-        f"/api/v1/devices",
+        f"/ui/bff/devices",
```
tsagadar, line 11: Could be a but cleaner directly updating the device model - so that this test module has no dependency to the bff api. Optional.

tsagadar (changes requested):

easybe:
> Commit [f674a68](https://github.com/UpstreamDataInc/goosebit/commit/f674a68993276a519486495ef501dfc9ac6ceee8) : either rewrite commit message (detail: no "." at the end of the first line) or ideally split in two (needs some git voodoo skills that pay off over time - or a good IDE).

Yes, that commit looks like a squash gone wrong. Better split it up if the 2 changes are unrelated.

This can be done with interactive rebase:
1. `git rebase -i origin/master`
1. Change first line to:
    ```
    e f674a68 # This is a combination of 2 commits. # This is the 1st commit message:
    ```
1. Save & quit
1. `git reset HEAD^` (or only extract some file(s) and amend)
1. Redo the two commits
1. `git rebase --continue`

Concerning the commit messages quality, there has been quite some progress :pray:. To take it one step further, I would appreciate if we could agree to follow the very common conventions outlined [here](https://cbea.ms/git-commit/), i.e.:
```
Summarize changes in around 50 characters (no period here)

More detailed explanatory text, if necessary. Wrap it to about 72
characters or so. In some contexts, the first line is treated as the
subject of the commit and the rest of the text as the body. The
blank line separating the summary from the body is critical (unless
you omit the body entirely); various tools like `log`, `shortlog`
and `rebase` can get confused if you run the two together.

Explain the problem that this commit is solving. Focus on why you
are making this change as opposed to how (the code explains that).
Are there side effects or other unintuitive consequences of this
change? Here's the place to explain them.

Further paragraphs come after blank lines.

 - Bullet points are okay, too

 - Typically a hyphen or asterisk is used for the bullet, preceded
   by a single space, with blank lines in between, but conventions
   vary here

If you use an issue tracker, put references to them at the bottom,
like this:

Closes: #123
```

tests/updater/controller/v1/test_routes.py
b-rowan: This seems a bit complicated just because of the way that endpoint functions, I think it introduces a lot of complexity directly into the tests that we may not want.  I think this should probably be left for a later PR in any case.

b-rowan:
> Yes, that commit looks like a squash gone wrong. Better split it up if the 2 changes are unrelated.

Yeah, should be fixed now.  I think they are related closely enough to be 1 commit (hence the failed squash).

tsagadar (approved):

PR #101, opened 2024-08-28
Reactivate aerich

### LINKED ISSUE #48: Database migrations
https://github.com/UpstreamDataInc/goosebit/blob/21975c0f5cc2dc678f84453225b91f1ada69703a/.gitignore#L168

Shouldn't migrations be committed? Am I missing something here?

Unfortunately, the commit message of 6e7c328ed9a2d45e84eda0e07b850d308ed12f5b is not very informative :disappointed:

tsagadar:
Giving aerich another shot - as we need a database migration mechanism for proper deployments.

Mainly removes the automated aerich tasks from application startup and leaves these steps to be
triggered by the developer or ops person (see also commit messages)

Let's see if upcoming model changes can be expressed by aerich as migrations.

Fixes #48


goosebit/settings/schema.py
```diff
@@ -63,7 +63,7 @@ def settings_customise_sources(
     ) -> tuple[PydanticBaseSettingsSource, ...]:
         return (
             init_settings,
-            YamlConfigSettingsSource(settings_cls, BASE_DIR.parent.joinpath("settings.yaml")),
+            YamlConfigSettingsSource(settings_cls, BASE_DIR.joinpath("settings.yaml")),
```
easybe, line 66: We have to keep in mind that we are creating a PyPI package. With that installed, this will point to e.g. `/usr/local/lib/python3.12/site-packages`, which is definitely not a good location. Anyway, this was already broken before.

easybe: I think instead of building this path ourselves, we should look into how to do this with a library.

goosebit/db/db.py
```diff
@@ -0,0 +1,15 @@
+from tortoise import Tortoise, run_async
```
b-rowan, line 1: I think this should be put in `__init__.py`?  Thoughts?

goosebit/settings/schema.py
b-rowan: Could also just allow the user to pass it as an arg to `main.py`, simple enough?

tsagadar: Created issue #103

goosebit/db/db.py
b-rowan: Not what I meant, I mean like I would prefer to just `from goosebit.db import init` rather than `from goosebit.db.db import init` or `from goosebit.db import db`, that seems strange to have a file structure like that.

Basically: rename `db.py` to `__init__.py`.

b-rowan: Regarding the commit also, that `init()` call wasn't entirely redundant, it was so you could run that file manually and have it attempt to init the DB (mostly for testing/inital startup).  Probably fine to remove it now though.

b-rowan (approved):

goosebit/db/db.py
tsagadar: Renamed as proposed.

Issue #103, opened 2024-08-29
Fix storage location of settings.yaml 


tsagadar:
When deployed using pip, the settings.yaml is searched within the `site-packages` folder structure (see https://github.com/UpstreamDataInc/goosebit/pull/101#discussion_r1734619755).

Also should review all the other usages of `BASE_DIR`

PR #105, opened 2024-08-29
Remove `BASE_DIR` parsing for settings

b-rowan:
`settings.yaml` being parsed from a base directory has issues when being installed from pip.  The user must now pass a `settings_file` path to goosebit when running.

With these changes, gooseBit should be run with `python -m goosebit [settings_file]`.  There is also now help information with `python -m goosebit --help`.

Fixes: #103


b-rowan:
Also note, `settings_file` has to be optional, because tests will break otherwise.  All settings should have a valid default.

goosebit/settings/schema.py
```diff
@@ -61,9 +63,17 @@ def settings_customise_sources(
...
     ) -> tuple[PydanticBaseSettingsSource, ...]:
+        # skip arg parsing if running pytest
+        if "pytest" not in sys.modules:
+            cli_args = parser.parse_args()
+            return (
+                init_settings,
+                YamlConfigSettingsSource(settings_cls, cli_args.settings_file),
+                env_settings,
+                file_secret_settings,
+            )
```
b-rowan, line 74: Not a fan of this personally, but I cant think of a better option with the current project structure...

goosebit/settings/args.py
```diff
@@ -0,0 +1,5 @@
+import argparse
```
easybe, line 1: This is not where I would expect argument parsing. IMHO this should be in `__main__.py`.

easybe (commented):
I don't think this is going into the right direction. What if you want to run the service with e.g. `gunicorn` together with a settings file?

I would rather see the app search for a system-wide file (e.g. in `/etc`) and maybe the current directory. This can be made portable with e.g. the `platformdirs` package.

goosebit/settings/schema.py
easybe: This looks awful.

b-rowan:
> I don't think this is going into the right direction. What if you want to run the service with e.g. `gunicorn` together with a settings file?
> 
> I would rather see the app search for a system-wide file (e.g. in `/etc`) and maybe the current directory. This can be made portable with e.g. the `platformdirs` package.

Fair enough.  Closing because I dont think this makes much sense then

PR #106, opened 2024-08-29
Better settings parsing and pip installation handling

b-rowan:
Hopefully a better solution to settings parsing than #105, uses `platformdirs` to find device specific directories where possible.  Will also log a warning about missing settings file, but will still check in project root for development.

Also moved `main.py` to `__main__.py` for better pip installation support, so it can be run with `python -m goosebit`.

Fixes: #103


goosebit/settings/const.py
```diff
@@ -1,8 +1,11 @@
 from pathlib import Path
 
+import platformdirs
 from argon2 import PasswordHasher
 
 BASE_DIR = Path(__file__).resolve().parent.parent.parent
+CONFIG_DIR = platformdirs.site_config_path("gooseBit")
```
easybe, line 7: Not a fan :disappointed:
```
>>> platformdirs.site_config_path("gooseBit")
PosixPath('/etc/xdg/xdg-ubuntu/gooseBit')
```

Is it not possible to just get `/etc/goosebit.yaml` or maybe `/etc/goosebit/settings.yaml`?

goosebit/settings/const.py
```diff
@@ -1,8 +1,11 @@
...
+DATA_DIR = platformdirs.site_data_path("gooseBit")
```
easybe, line 8:
```
>>> platformdirs.site_data_path("gooseBit")
PosixPath('/usr/share/ubuntu/gooseBit')
```
Better, but, I would expect:
```
/usr/share/goosebit
```

Also, the dev use case (as it was before) should still be covered. You don't want to run the app as `root`.

easybe (changes requested):

goosebit/__main__.py
```diff
@@ -0,0 +1,12 @@
...
+import uvicorn
+
+from goosebit import app
+from goosebit.settings import config
+
+logging.config.dictConfig(config.logging)
+
+
+if __name__ == "__main__":
+    uvicorn.run(app, host="0.0.0.0", port=config.port)
```
easybe, line 12: So `main.py` can be removed, no?

b-rowan: It can be, we may want to keep it for testing, but it is essentially a duplicate now.

goosebit/settings/const.py
b-rowan: Not sure.  I don't know why platformdirs wants these weird paths, I'll look into what other options we have there.

goosebit/settings/const.py
b-rowan: Not sure how to cover the dev use case here.  No way to know if we're installed from pip vs in a dev environment, technically you can just run in dev by specifying the DB and migrations locations manually, but that's a bit annoying.

goosebit/settings/const.py
b-rowan: I think the only other option for this might be `user_config_path`, I checked every other platformdirs call, and none of them look any better.  This resolves to `~/.config/gooseBit` when called with `user_config_path("gooseBit)` for me.

goosebit/settings/const.py
easybe: The solution is to look in multiple places. Many applications do it that way.

b-rowan: > The solution is to look in multiple places. Many applications do it that way.

We want to write to `DATA_DIR`, do we have a preference for location of that?

easybe: For writing the migrations, only the dev use case applies, so `<project root>/goosebit/db/migrations/` seems fine to me. And for reading, there where they are now, in the package directory, is fine too. I mean, it is even Python code.

goosebit/settings/const.py
easybe: On Linux/Unix, searching these locations in the following order would be fine:
  1. `settings.yaml` in `$PWD` (or relative to the code) for development and people running it from e.g. `/app` in a container
  1. `~/.config/goosebit/settings.yaml` for a root-less installation
  1. `/etc/goosebit/settings.yaml` for a system-wide installation

For Windows, the first one should already be enough, so we could also just hardcode those.

b-rowan: > For Windows, the first one should already be enough, so we could also just hardcode those.

I think we found out with the whole crowdstrike incident how many people are using Windows for things they shouldn't really be, so I'm not certain how much we want to support Windows.  This is essentially a breaking change for using it with `pip` in windows, which may be annoying if trying to test.

Overall up to you, I like the directories you listed, but I am generally of the opinion that trying to support all major systems as well as possible is a good idea.

goosebit/settings/const.py
b-rowan: > For writing the migrations, only the dev use case applies, so `<project root>/goosebit/db/migrations/` seems fine to me. And for reading, there where they are now, in the package directory, is fine too. I mean, it is even Python code.

Works for me.  And the user can already set the DB location in the config, so thats fine.  Don't need a DATA_DIR.

easybe: Yes, same goes for `artifacts`. The current default is fine. When installed with pip, the default will not be writable by the application (unless run as root), so the user needs to configure the path. A little ugly, but it should work.

goosebit/settings/const.py
easybe: What I tried to say is, if we support loading `settings.yaml` from the current working directory, a clean deployment can be achieved on Windows. No need to search `%AppData%` or whatever.

On second thought, why not try `os.getcwd()` and then fallback to the `GOOSEBIT_SETTINGS` environment variable. That keeps it simple and flexible at the same time.

b-rowan: > What I tried to say is, if we support loading `settings.yaml` from the current working directory, a clean deployment can be achieved on Windows. No need to search `%AppData%` or whatever.
> 
> On second thought, why not try `os.getcwd()` and then fallback to the `GOOSEBIT_SETTINGS` environment variable. That keeps it simple and flexible at the same time.

Yeah this seems like the best solution.  The only problem is we are loading our environment variables with the same loader that loads the yaml file, but I think I can make another one, load it, then use it for the other one.

Regarding directory default, do you think going `os.cwd()`, then goosebit root, then `GOOSEBIT_SETTINGS` will work best?  Or do we just ignore the root and hope that development envs are running it properly?

easybe: > Yeah this seems like the best solution. The only problem is we are loading our environment variables with the same loader that loads the yaml file, but I think I can make another one, load it, then use it for the other one.

Just do a `os.getenv()` for that one.

> Regarding directory default, do you think going `os.cwd()`, then goosebit root, then `GOOSEBIT_SETTINGS` will work best? Or do we just ignore the root and hope that development envs are running it properly?

I guess also checking the project root doesn’t hurt.

b-rowan: Sounds good.  Going to bed for now, but will take another stab at this tomorrow.

goosebit/__main__.py
b-rowan: I'm going to leave it in for now, but can be removed later if we decide to.

goosebit/settings/const.py
```diff
@@ -1,8 +1,11 @@
 from pathlib import Path
 
+import platformdirs
```
easybe: Introducing this dependency just to be removed later is confusing. I think some squashing is in order.

b-rowan: Squashed.

goosebit/settings/schema.py
```diff
@@ -46,7 +51,8 @@ class GooseBitSettings(BaseSettings):
 
     users: list[User] = []
 
-    db_uri: str = f"sqlite:///{BASE_DIR.joinpath('db.sqlite3')}"
+    db_uri: str = f"sqlite:///{GOOSEBIT_ROOT_DIR.joinpath('db.sqlite3')}"
```
easybe, line 54: I forgot about this one, not a great default value in production, where the package is installed system-wide.
Here's an idea:
```python
project_root = Path(__file__).resolve().parent.parent.parent

if not (project_root / "pyproject.toml").exists():
    project_root = None

if project_root is not None:
    db_uri: str = f"sqlite:///{project_root / 'db.sqlite3'}"
else:
    db_uri: str = f"sqlite:////var/lib/goosebit/db.sqlite3'"   # I guess we need to make sure the dir exists
```

goosebit/settings/schema.py
```diff
@@ -61,9 +67,23 @@ def settings_customise_sources(
...
-            env_settings,
-            file_secret_settings,
+            YamlConfigSettingsSource(settings_cls, GOOSEBIT_ROOT_DIR.joinpath("goosebit.yaml")),
+            YamlConfigSettingsSource(settings_cls, CURRENT_DIR.joinpath("goosebit.yaml")),
+        ]
+        if ETC_SETTINGS_DIR.exists():
+            settings_sources.append(
+                YamlConfigSettingsSource(settings_cls, ETC_SETTINGS_DIR.joinpath("goosebit.yaml")),
+            )
+        if os.getenv("GOOSEBIT_SETTINGS"):
```
easybe, line 79:
```suggestion
        if os.getenv("GOOSEBIT_SETTINGS") is not None:
```

Anyway, IMHO the env var should have precedence.

easybe (commented):
And of course some nagging about the commit messages. I appreciate that they now contain more information. But, reading them in the console is not that great:

Screenshot from 2024-08-31 23-02-42

That is why one should break the lines, as explained [here](https://cbea.ms/git-commit/). A common guide line is 50 characters for the subject and 72 for the body.

So, your commit I would change to something like this:
```
Check multiple places for settings file

Also, rename settings.yaml to `goosebit.yaml` to make it distinguishable
from other files.

Use can now pass `GOOSEBIT_SETTINGS` to customize settings location, as
well as gooseBit auto checking `/etc/goosebit.yaml`, the root of the
project directory, and the current working directory when it is run.
This should allow for more optionality when loading settings files.
    
Fixes: #103
```

b-rowan:
> And of course some nagging about the commit messages. I appreciate that they now contain more information. But, reading them in the console is not that great:
> 
> 
> 
> Screenshot from 2024-08-31 23-02-42
> 
> 
> 
> That is why one should break the lines, as explained [here](https://cbea.ms/git-commit/). A common guide line is 50 characters for the subject and 72 for the body. Pro tip: Editors can help you with that.
> 
> 
> 
> So, your commit message I would change to something like this:
> 
> ```
> 
> Check multiple places for settings file
> 
> 
> 
> Also, rename settings.yaml to `goosebit.yaml` to make it distinguishable
> 
> from other files.
> 
> 
> 
> Use can now pass `GOOSEBIT_SETTINGS` to customize settings location, as
> 
> well as gooseBit auto checking `/etc/goosebit.yaml`, the root of the
> 
> project directory, and the current working directory when it is run.
> 
> This should allow for more optionality when loading settings files.
> 
>     
> 
> Fixes: #103
> 
> ```
> 
> 
> 
>  
> 
> 

I wonder if this is a good case for pre-commit?

easybe:
> > And of course some nagging about the commit messages. I appreciate that they now contain more information. But, reading them in the console is not that great:
> > Screenshot from 2024-08-31 23-02-42
> > That is why one should break the lines, as explained [here](https://cbea.ms/git-commit/). A common guide line is 50 characters for the subject and 72 for the body. Pro tip: Editors can help you with that.
> > So, your commit message I would change to something like this:
> > ```
> > 
> > Check multiple places for settings file
> > 
> > 
> > 
> > Also, rename settings.yaml to `goosebit.yaml` to make it distinguishable
> > 
> > from other files.
> > 
> > 
> > 
> > Use can now pass `GOOSEBIT_SETTINGS` to customize settings location, as
> > 
> > well as gooseBit auto checking `/etc/goosebit.yaml`, the root of the
> > 
> > project directory, and the current working directory when it is run.
> > 
> > This should allow for more optionality when loading settings files.
> > 
> >     
> > 
> > Fixes: #103
> > ```
> 
> I wonder if this is a good case for pre-commit?

There is https://github.com/jorisroovers/gitlint

goosebit/settings/schema.py
b-rowan: I can't say I'm a fan of this.  Container should use postgres nearly always, and IMO the user should be setting a DB location in every case.  This really should only be used in testing.

If we care that much about this, we can warn about defaults.

goosebit/settings/schema.py
b-rowan: > Anyway, IMHO the env var should have precedence.

My understanding was that it does.  Seems like these are loaded first to last, and later values can overwrite earlier ones.

goosebit/settings/schema.py
b-rowan: I think the case is to either `warnings.warn` if the DB is on default and the `pyproject.toml` doesn't exist, OR to outright stop the program with an error.  Don't even give the user an option to run it wrong.

b-rowan:
> > > And of course some nagging about the commit messages. I appreciate that they now contain more information. But, reading them in the console is not that great:
> > > Screenshot from 2024-08-31 23-02-42
> > > That is why one should break the lines, as explained [here](https://cbea.ms/git-commit/). A common guide line is 50 characters for the subject and 72 for the body. Pro tip: Editors can help you with that.
> > > So, your commit message I would change to something like this:
> > > ```
> > > 
> > > Check multiple places for settings file
> > > 
> > > 
> > > 
> > > Also, rename settings.yaml to `goosebit.yaml` to make it distinguishable
> > > 
> > > from other files.
> > > 
> > > 
> > > 
> > > Use can now pass `GOOSEBIT_SETTINGS` to customize settings location, as
> > > 
> > > well as gooseBit auto checking `/etc/goosebit.yaml`, the root of the
> > > 
> > > project directory, and the current working directory when it is run.
> > > 
> > > This should allow for more optionality when loading settings files.
> > > 
> > >     
> > > 
> > > Fixes: #103
> > > ```
> > 
> > 
> > I wonder if this is a good case for pre-commit?
> 
> There is https://github.com/jorisroovers/gitlint

#110

easybe:
I'll want to test this...

goosebit/settings/schema.py
```diff
@@ -61,9 +67,23 @@ def settings_customise_sources(
...
+                YamlConfigSettingsSource(settings_cls, ETC_SETTINGS_FILE),
+            )
+        if os.getenv("GOOSEBIT_SETTINGS") is not None:
+            settings_sources.append(
+                YamlConfigSettingsSource(settings_cls, os.getenv("GOOSEBIT_SETTINGS")),
+            )
+        settings_sources.extend(
+            [
+                env_settings,
+                file_secret_settings,
```
easybe, line 86: We don't use secrets, so we can remove this, no?

goosebit/settings/schema.py
```diff
@@ -61,9 +67,23 @@ def settings_customise_sources(
         dotenv_settings: PydanticBaseSettingsSource,
         file_secret_settings: PydanticBaseSettingsSource,
     ) -> tuple[PydanticBaseSettingsSource, ...]:
-        return (
+        settings_sources = [
             init_settings,
```
easybe, line 71: Do you know where `init_settings` comes from? As it is at the top, I should have the highest priority. Not sure if that is what we want.

easybe (changes requested):

goosebit/settings/schema.py
b-rowan: I think you have it in the wrong direction still?  Init settings should be default settings I assume.

b-rowan: Notice how `env_settings`  comes after the yaml file handler, as ENV always has precedence over the yaml files.  Same thing here, because `init_settings` is first in the tuple, it is the first to get parsed and can then be overwritten.

goosebit/settings/schema.py
b-rowan: Sure.  The pydantic example I saw seems to want to include those always, but fine to remove.

goosebit/settings/schema.py
```diff
@@ -61,9 +64,32 @@ def settings_customise_sources(
...
-            YamlConfigSettingsSource(settings_cls, BASE_DIR.joinpath("settings.yaml")),
-            env_settings,
-            file_secret_settings,
+            YamlConfigSettingsSource(settings_cls, GOOSEBIT_ROOT_DIR.joinpath("goosebit.yaml")),
```
easybe, line 69: These two lines are now redundant and should be removed.

goosebit/settings/schema.py
```diff
@@ -61,9 +64,32 @@ def settings_customise_sources(
...
+
+        for config_file in config_files:
+            if config_file.exists():
+                settings_sources.append(
+                    YamlConfigSettingsSource(settings_cls, config_file),
+                )
+                logger.info(f"Selected config file to load: {config_file}")
+                break
+
+        settings_sources.append(env_settings)
```
easybe, line 94: Environment variables should should override everything as they are the most explicit.

goosebit/settings/schema.py
easybe: I am a little confused because the docs state

> The order of the returned callables decides the priority of inputs; first item is the highest priority.

https://docs.pydantic.dev/latest/concepts/pydantic_settings/#customise-settings-sources

easybe (changes requested):

goosebit/settings/schema.py
b-rowan: > Do you know where init_settings comes from? As it is at the top, it should have the highest priority. Not sure if that is what we want.

Seems like these are settings passed to `__init__`, so can remove that as well.

goosebit/settings/schema.py
```diff
@@ -61,9 +64,25 @@ def settings_customise_sources(
...
+                Path("/etc/goosebit.yaml"),
+            ]
         )
+
+        for config_file in config_files:
+            if config_file.exists():
+                settings_sources.append(
+                    YamlConfigSettingsSource(settings_cls, config_file),
+                )
+                logger.info(f"Selected config file to load: {config_file}")
```
easybe, line 86: This doesn't seem to work. I think at this point we cannot log yet.

b-rowan: Yeah, this is run before the logger is instantiated.  Might need a new PR to modify the way the logger works, move it to another file so it can just log that after the logger is loaded.

easybe: OK. In the meantime, I would just remove this and then this PR is good to go, I guess.

b-rowan: I actually just added it, didn't notice that message.  I can remove it if you want, but its a good look at what needs to be done to accomplish it.

goosebit/settings/schema.py
```diff
@@ -61,9 +64,25 @@ def settings_customise_sources(
...
+
+        config_files.extend(
+            [
+                CURRENT_DIR.joinpath("goosebit.yaml"),
+                GOOSEBIT_ROOT_DIR.joinpath("goosebit.yaml"),
+                Path("/etc/goosebit.yaml"),
+            ]
         )
+
+        for config_file in config_files:
```
easybe, line 81:
```suggestion
        cls.config_file = None
        for config_file in config_files:
```

easybe (changes requested):
And I would squash the last commit. No need for intermediate (broken) versions.

goosebit/settings/schema.py
easybe: Or do this in constructor.

b-rowan:
> And I would squash the last commit. No need for intermediate (broken) versions.

Will do, just give me a bit.

easybe (approved):
This is a great improvement, thanks!

PR #108, opened 2024-08-30
Add initial DB migration for PostgreSQL

easybe:
PostgreSQL does not support AUTOINCREMENT. Instead it has the SERIAL type.

To generate the migration `aerich init-db` together with PostgreSQL 16.3 was used.


tsagadar (approved):

goosebit/db/migrations/models/0_20240830054046_init_postgres.py
```diff
@@ -0,0 +1,75 @@
+from tortoise import BaseDBAsyncClient
```
tsagadar, line 1: This has to be merged with the initial migration: the very first migration has to create the aerich relation, so that aerich can start tracking the applied migrations. For Postgres, this is not the case if only the second migration creates the table.

tsagadar (changes requested):

goosebit/db/migrations/models/0_20240830054046_init_postgres.py
easybe: I see. That is how I had it in the beginning. Will do, thanks.

easybe:
OK, now it works.

tsagadar:
The better approach would be to update the initial migration than to delete it and to create a new one. aerich might not like having updates applied in a dev pc that no longer exist.

All dev dbs that have the initial migration applied don't need the postgres flavor.

easybe:
> The better approach would be to update the initial migration than to delete it and to create a new one. aerich might not like having updates applied in a dev pc that no longer exist.
> 
> All dev dbs that have the initial migration applied don't need the postgres flavor.

I don’t understand. Depending on the DB different migration code is used. Sounds perfect to me.

tsagadar:
The aerich table stores the full file name. Not going to fully reverse engineer aerich to figure out if we play with the fire or not by renaming migration files while keeping the initial version. So ok for me.

tsagadar (approved):

b-rowan:
Hate to be this guy after this is already ready, but is it possible we can support multiple databases in migrations by changing the tortoise config?  EG here - https://tortoise.github.io/setup.html#tortoise.Tortoise.init-parameters

They seem to show that you can have `default` db config, as well as possibly others, and choose between them.  I wonder if it would improve migrations if we added a `postgres` and `sqlite` DB in there (even if unused), because it would force aerich to generate schemas for both?

easybe:
> Hate to be this guy after this is already ready, but is it possible we can support multiple databases in migrations by changing the tortoise config? EG here - https://tortoise.github.io/setup.html#tortoise.Tortoise.init-parameters
> 
> They seem to show that you can have `default` db config, as well as possibly others, and choose between them. I wonder if it would improve migrations if we added a `postgres` and `sqlite` DB in there (even if unused), because it would force aerich to generate schemas for both?

I doubt that we can trick aerich in doing that. A quick try supports that assumption. Although aerich seems to support multiple DBs, you have to associate models to them: https://github.com/tortoise/aerich?tab=readme-ov-file#multiple-databases

b-rowan:
> > Hate to be this guy after this is already ready, but is it possible we can support multiple databases in migrations by changing the tortoise config? EG here - https://tortoise.github.io/setup.html#tortoise.Tortoise.init-parameters
> > They seem to show that you can have `default` db config, as well as possibly others, and choose between them. I wonder if it would improve migrations if we added a `postgres` and `sqlite` DB in there (even if unused), because it would force aerich to generate schemas for both?
> 
> I doubt that we can trick aerich in doing that. A quick try supports that assumption. Although aerich seems to support multiple DBs, you have to associate models to them: https://github.com/tortoise/aerich?tab=readme-ov-file#multiple-databases

Makes sense.  Mildly annoying honestly, not sure why it doesn't seem to allow you to chose DBs (or even generate for all DBs)...

easybe:
> > > Hate to be this guy after this is already ready, but is it possible we can support multiple databases in migrations by changing the tortoise config? EG here - https://tortoise.github.io/setup.html#tortoise.Tortoise.init-parameters
> > > They seem to show that you can have `default` db config, as well as possibly others, and choose between them. I wonder if it would improve migrations if we added a `postgres` and `sqlite` DB in there (even if unused), because it would force aerich to generate schemas for both?
> > 
> > 
> > I doubt that we can trick aerich in doing that. A quick try supports that assumption. Although aerich seems to support multiple DBs, you have to associate models to them: https://github.com/tortoise/aerich?tab=readme-ov-file#multiple-databases
> 
> Makes sense. Mildly annoying honestly, not sure why it doesn't seem to allow you to chose DBs (or even generate for all DBs)...

Well, it is an open source project, I bet they won’t be opposed to such contributions 😉

PR #113, opened 2024-09-03
Fixing software upload and return ID of created resource

### LINKED ISSUE #111: Post to `/api/v1/software` does not return ID
The ID of the software entry should be returned on successful creation. The ID is needed to create a corresponding rollout.

tsagadar:
Fully implemented POST /api/v1/software . Added tests as well and some minor cleanups.

Fixes #111


goosebit/api/v1/software/routes.py
```diff
@@ -52,7 +55,7 @@ async def software_delete(_: Request, delete_req: SoftwareDeleteRequest) -> Stat
     "",
     dependencies=[Security(validate_user_permissions, scopes=["software.write"])],
 )
-async def post_update(_: Request, file: UploadFile = File(None), url: str = Form(None)):
+async def post_update(_: Request, file: Optional[UploadFile] = File(None), url: Optional[str] = Form(None)):
```
b-rowan, line 58: I think we prefer `T | None` over `Optional[T]` generally.

b-rowan (changes requested):

goosebit/api/v1/software/routes.py
tsagadar: Cleaned up this code and the other places where Optional was used.

b-rowan (approved):

PR #115, opened 2024-09-03
Auto assign milestone to PR or issue based on tag

b-rowan:
Assigns a milestone to an issue or PR automatically when a tag matching a `v*.*.*` semantic version is added.

### DIFF
--- a/.github/workflows/issue-milestone-on-tag-remove.yml
+++ b/.github/workflows/issue-milestone-on-tag-remove.yml
@@ -1,0 +1,22 @@
+on:
+  issues:
+    types: [unlabeled]
+
+jobs:
+  remove-milestone:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Check if tag is a valid version
+        uses: actions-ecosystem/action-regex-match@v2
+        id: regex-match
+        with:
+          text: ${{ github.event.label.name }}
+          regex: 'v[0-9]+?\.[0-9]+?\.[0-9]+?'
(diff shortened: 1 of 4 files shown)


tsagadar:
I am not that familiar with this. My current understanding is that these actions will automatically assign/unassign issues/PRs to milestones if a label is used matching a version? Shouldn't we just directly set those milestones instead of labels (personally I was not that aware about milestones, but sure the better approach).

What confused me initially was the term "tag". So I assumed git tags. Maybe better to use label?

b-rowan:
> I am not that familiar with this. My current understanding is that these actions will automatically assign/unassign issues/PRs to milestones if a label is used matching a version? Shouldn't we just directly set those milestones instead of labels (personally I was not that aware about milestones, but sure the better approach).

I sort of agree, but the tagging system is a bit nicer to work with I think.  Basically the goal of this PR is to use tags to synchronize with milestones, as milestones seem to track progress better, but tags are easier to assign and work with.  We create the tags and then the rest gets automated.

If the consensus is a preference of milestone only for versions, I'm ok with that too, but this actually seemed less confusing to me.

> What confused me initially was the term "tag". So I assumed git tags. Maybe better to use label?

Yeah, this makes sense.  I was a bit confused too when I wrote this, I'll swap everything over to labels tomorrow.

easybe:
> If the consensus is a preference of milestone only for versions, I'm ok with that too, but this actually seemed less confusing to me.

Yes, we should use milestones instead of labels for versions, because that is what the feature is for. Even though I was aware of milestones, I started using labels because it was the easiest at the time.

IMHO we should only introduce a new CI job (and complexity) if there is an actual problem to solve.

Issue #118, opened 2024-09-04
Docker build fails on release


b-rowan:
See log - https://github.com/UpstreamDataInc/goosebit/actions/runs/10709345170/job/29693758319

easybe:
I assume `GITHUB_REF_TYPE` is not set to `tag` because the workflow is triggered by the completion of the `PyPI` workflow and no tag push. I will have to look into that tomorrow.

PR #119, opened 2024-09-04
Fix Docker CI workflow

easybe:
As we are triggered by the `PyPI` workflow and not directly by the push of the Git tag, extracting the version from `GITHUB_REF_NAME` won't work. Calling `git describe` instead also allows the workflow to be run manually, which makes things easier in general.

Fixes: #118


easybe:
Test run: https://github.com/UpstreamDataInc/goosebit/actions/runs/10710072820

b-rowan (approved):

Issue #104, opened 2024-08-29
Support different (local) software with same file name


tsagadar:
Currently the uploaded software files have to be named differently to avoid a name clash.

Issue and solution approach discussed here https://github.com/UpstreamDataInc/goosebit/issues/81#issuecomment-2315597149

PR #123, opened 2024-09-05
Support local files with matching filenames

b-rowan:
Adds support for local files with matching filenames but different hashes.  This is done by moving the path of the files to `<artifact_dir>/<file_hash>/<filename>`.  Files with matching versions and compatibility will still collide and raise an error to the user.

Fixes: #104

### DIFF
--- a/goosebit/db/__init__.py
+++ b/goosebit/db/__init__.py
@@ -1,11 +1,21 @@
+from logging import getLogger
+
 from tortoise import Tortoise
+from tortoise.exceptions import OperationalError
 
 from goosebit.db.config import TORTOISE_CONF
+from goosebit.db.models import Device
+
+logger = getLogger(__name__)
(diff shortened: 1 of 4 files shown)


goosebit/db/__init__.py
```diff
@@ -5,6 +5,7 @@
 
 async def init():
     await Tortoise.init(config=TORTOISE_CONF)
+    await Tortoise.generate_schemas()
```
tsagadar, line 8: Should not be necessary. As written in the README.md : "Create or upgrade database: poetry run aerich upgrade"

My working assumption is, that DB manipulations are best left to the Dev / Ops person. Don't automate it within the app.

tsagadar (changes requested):

tsagadar (approved):

PR #126, opened 2024-09-07
Migrate file operations to async

### LINKED ISSUE #124: Use async methods for all file operations
fastapi as an ASGI server works best if the endpoints solely rely on async functions when interacting with external resources (database, web servers, files).

A quick search for "file" revealed a few spots where synchronous calls are used. The following list of functions should help to quickly identify those spots:
- `file_digest`
- `stat`
- `copy`
- `unlink`
- `mkdir`

Note 1: there could be other synchronous calls in our code. Not a big problem if a few are missed as fastapi uses thread pool to execute those
Note 2: endpoints should all be async so that they can be executed in parallel. This is currently ok - unless for those that do pure "computing" and don't use external resources.

Detailled technical information in first answer to this [StackOverflow question](https://stackoverflow.com/questions/71516140/fastapi-runs-api-calls-in-serial-instead-of-parallel-fashion?noredirect=1&lq=1).

tsagadar:
More for consistency than for improved performance (as aiofiles also uses threads to execute synchronous functions like fastapi would).

Also fixes an issue deleting software: containing folder was not removed.

Fixes #124

### DIFF
--- a/goosebit/api/v1/software/routes.py
+++ b/goosebit/api/v1/software/routes.py
@@ -1,6 +1,7 @@
 from pathlib import Path
 
 import aiofiles
+from aiofiles import os
 from fastapi import APIRouter, File, Form, HTTPException, Security, UploadFile
 from fastapi.requests import Request
 
@@ -41,12 +42,14 @@ async def software_delete(_: Request, delete_req: SoftwareDeleteRequest) -> Stat
             raise HTTPException(409, "Software is referenced by rollout")
 
         if software.local:
-            path = software.path
-            if path.exists():
-                path.unlink()
+            try:
+                await os.unlink(software.path)
+            except OSError:
+                pass
 
         await software.delete()
         success = True
+
     return StatusResponse(success=success)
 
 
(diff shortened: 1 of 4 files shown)


b-rowan (approved):

jameshilliard:
I think fastapi uses [anyio](https://anyio.readthedocs.io/en/stable/fileio.html) instead of aiofiles, I wonder if that may be better to use than aiofiles here as well.

jameshilliard:
Made a anyio based version of this in #127.

b-rowan:
I'm leaning towards anyio for file handling, I think removing aiofiles as a dependency is a good idea to keep the dependencies as minimal as possible.

easybe:
> I'm leaning towards anyio for file handling, I think removing aiofiles as a dependency is a good idea to keep the dependencies as minimal as possible.

Is there a replacement for `aiofiles.tempfile`?

tsagadar:
Closing in favor of better approach in #127

PR #127, opened 2024-09-07
Use anyio based async file io operations

jameshilliard:
Alternative to #126.

### DIFF
--- a/goosebit/ui/bff/software/routes.py
+++ b/goosebit/ui/bff/software/routes.py
@@ -1,6 +1,6 @@
 from __future__ import annotations
 
-import aiofiles
+from anyio import Path, open_file
 from fastapi import APIRouter, Form, HTTPException, Security, UploadFile
 from fastapi.requests import Request
 from tortoise.expressions import Q
@@ -64,20 +64,20 @@ async def post_update(
         await create_software_update(url, None)
     else:
         # local file
-        file = config.artifacts_dir.joinpath(filename)
-        config.artifacts_dir.mkdir(parents=True, exist_ok=True)
+        artifacts_dir = Path(config.artifacts_dir)
+        file = artifacts_dir.joinpath(filename)
+        await artifacts_dir.mkdir(parents=True, exist_ok=True)
 
         temp_file = file.with_suffix(".tmp")
         if init:
-            temp_file.unlink(missing_ok=True)
+            await temp_file.unlink(missing_ok=True)
 
-        contents = await chunk.read()
-
-        async with aiofiles.open(temp_file, mode="ab") as f:
-            await f.write(contents)
+        async with await open_file(temp_file, "ab") as f:
+            await f.write(await chunk.read())
 
         if done:
             try:
-                await create_software_update(file.absolute().as_uri(), temp_file)
+                absolute = await file.absolute()
+                await create_software_update(absolute.as_uri(), temp_file)
             finally:
-                temp_file.unlink(missing_ok=True)
+                await temp_file.unlink(missing_ok=True)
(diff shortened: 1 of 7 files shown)


goosebit/ui/bff/software/routes.py
```diff
@@ -64,20 +64,20 @@ async def post_update(
...
         temp_file = file.with_suffix(".tmp")
         if init:
-            temp_file.unlink(missing_ok=True)
+            await temp_file.unlink(missing_ok=True)
 
-        contents = await chunk.read()
-
-        async with aiofiles.open(temp_file, mode="ab") as f:
-            await f.write(contents)
+        async with await open_file(temp_file, "ab") as f:
```
b-rowan, line 75: I think we should confirm a preference between aiofiles and anyio for file handling across the project.  I'm alright with using anyio for file handling, but in that case we should work towards complete removal of aiofiles.

goosebit/updates/swdesc.py
```diff
@@ -72,7 +73,17 @@ async def parse_remote(url: str):
...
+    sha1_hash = hashlib.sha1()
+    buf = bytearray(2**18)
+    view = memoryview(buf)
+    while True:
+        size = await fileobj.readinto(buf)
+        if size == 0:
+            break
+        sha1_hash.update(view[:size])
+
+    await fileobj.seek(last)
```
b-rowan, line 88: Is there a reason to do this in this format?  Probably better on RAM, but even the worst server equipment should have enough ram to handle a few of these files at a time...

tests/api/v1/software/test_routes.py
```diff
@@ -1,13 +1,13 @@
...
 @pytest.mark.asyncio
 async def test_create_software_local(async_client, test_data):
-    path = Path(__file__).resolve().parent / "software-header.swu"
-    with open(path, "rb") as file:
-        files = {"file": file}
+    resolved = await Path(__file__).resolve()
+    path = resolved.parent / "software-header.swu"
+    async with await open_file(path, "rb") as file:
+        files = {"file": await file.read()}
         response = await async_client.post(f"/api/v1/software", files=files)
```
b-rowan, line 11: This should drop back a level to be outside the scope of the `async with`.  Not introduced here, but should be fixed nonetheless.

b-rowan (changes requested):
I prefer this over adding another dependency, and using anyio may in the future allow this to be run with different event loops, which I think is a positive.

Overall we should decide how we want file IO to be handled (as opposed to filesystem IO as handled in the PR).  I would like to propose that we remove all calls which use aiofiles (and remove the dependency on it) and replace them with anyio file IO.

goosebit/updates/swdesc.py
jameshilliard: This is basically a straight port of the stdlib's `hashlib.file_digest` function to use anyio instead of normal synchronous IO. IMO we should generally operate under the assumption that swu files may not fit into ram always, the upper limit on swu file sizes is rather high as well.

tsagadar:
I think it is a good idea to limit dependencies - was not aware that fastapi uses anyio. anyio also has 4x as many monthly downloads than aiofiles. So definitely a solid choice. So I closed #126.

To complete PR I'd also suggest to fully move away from aiofiles and replace the temp file handling as well.

jameshilliard:
> To complete PR I'd also suggest to fully move away from aiofiles and replace the temp file handling as well.

I didn't find temp file handling support in anyio unfortunately.

goosebit/ui/bff/software/routes.py
b-rowan: We can look at this in later versions, without tempfile support this is likely not possible yet.

b-rowan (approved):

PR #128, opened 2024-09-07
Enable mypy type checking

jameshilliard:
This seems to break some tests.

### DIFF
--- a/goosebit/auth/__init__.py
+++ b/goosebit/auth/__init__.py
@@ -27,15 +27,15 @@ def create_token(username: str) -> str:
     return jwt.encode(header={"alg": "HS256"}, claims={"username": username}, key=config.secret_key)
 
 
-def get_user_from_token(token: str) -> User | None:
+def get_user_from_token(token: str | None) -> User | None:
     if token is None:
-        return
+        return None
     try:
         token_data = jwt.decode(token, config.secret_key)
         username = token_data.claims["username"]
         return USERS.get(username)
     except (BadSignatureError, LookupError, ValueError):
-        pass
+        return None
 
 
 def login_user(username: str, password: str) -> str:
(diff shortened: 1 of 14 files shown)


goosebit/auth/__init__.py
```diff
@@ -58,17 +58,17 @@ def login_user(username: str, password: str) -> str:
 
 
 def get_current_user(
-    session_token: Annotated[str, Depends(session_auth)] = None,
-    oauth2_token: Annotated[str, Depends(oauth2_auth)] = None,
-) -> User:
+    session_token: Annotated[str, Depends(session_auth)] | None = None,
+    oauth2_token: Annotated[str, Depends(oauth2_auth)] | None = None,
```
b-rowan, line 62: Need to move type hint inside `Annotated`, should be -

```
def get_current_user(
    session_token: Annotated[str | None, Depends(session_auth)] = None,
    oauth2_token: Annotated[str | None, Depends(oauth2_auth)] = None,
) -> User:
```

b-rowan (changes requested):

goosebit/updater/manager.py
```diff
@@ -137,11 +138,16 @@ def __init__(self, dev_id: str):
...
+        device = await self.get_device()
+        if device is None:
+            return None
+        return await Software.latest(device)
 
-    async def get_update(self) -> tuple[HandlingType, Software]:
+    async def get_update(self) -> tuple[HandlingType, Software] | None:
         software = await self._get_software()
+        if software is None:
+            return None
```
b-rowan, line 150: I think this can cause issues, I would rather return `HandlingType.SKIP, None`, and have the type be `tuple[Software | None, HandlingType]`

goosebit/updater/manager.py
```diff
@@ -276,9 +282,11 @@ async def _get_software(self) -> Software | None:
         assert device.update_mode == UpdateModeEnum.PINNED
         return None
 
-    async def get_update(self) -> tuple[HandlingType, Software]:
+    async def get_update(self) -> tuple[HandlingType, Software] | None:
```
b-rowan, line 285: Same as above, would rather have `Software | None` rather than `tuple[...] | None`

goosebit/updates/__init__.py
```diff
@@ -90,7 +90,10 @@ async def _is_software_colliding(update_info):
 
 
 async def generate_chunk(request: Request, updater: UpdateManager) -> list:
-    _, software = await updater.get_update()
+    update = await updater.get_update()
+    if update is None:
+        return []
```
b-rowan, line 95: After applying the changes above, this needs to be changed.

b-rowan (changes requested):

b-rowan (approved):

PR #129, opened 2024-09-09
Use `model_validate` built into pydantic to parse DB models

b-rowan:
`pydantic` models have a deprecated `from_orm` method, which is designed to parse DB models into the schema. This was superseded by `model_validate`, which does the same thing when the schema config `from_attributes` value is true. Migrating to using this simplifies the method of parsing DB models, and will eventually allow the schemas to be used internally instead of the DB models.

Reference here - https://docs.pydantic.dev/2.0/usage/models/#arbitrary-class-instances

The goal here is to get to the point where we can pass around arbitrary abstract software types internally, which will allow hooking the software that a device should receive.

### DIFF
--- a/goosebit/api/v1/rollouts/routes.py
+++ b/goosebit/api/v1/rollouts/routes.py
@@ -1,9 +1,9 @@
-from fastapi import APIRouter, Security
+from fastapi import APIRouter, HTTPException, Security
 from fastapi.requests import Request
 
 from goosebit.api.responses import StatusResponse
 from goosebit.auth import validate_user_permissions
-from goosebit.db.models import Rollout
+from goosebit.db.models import Rollout, Software
 
 from .requests import RolloutsDeleteRequest, RolloutsPatchRequest, RolloutsPutRequest
 from .responses import RolloutsPutResponse, RolloutsResponse
@@ -24,6 +24,9 @@ async def rollouts_get(_: Request) -> RolloutsResponse:
     dependencies=[Security(validate_user_permissions, scopes=["rollout.write"])],
 )
 async def rollouts_put(_: Request, rollout: RolloutsPutRequest) -> RolloutsPutResponse:
+    software = await Software.filter(id=rollout.software_id)
+    if len(software) == 0:
+        raise HTTPException(404, f"No software with ID {rollout.software_id} found")
     rollout = await Rollout.create(
         name=rollout.name,
         feed=rollout.feed,
--- a/goosebit/api/v1/rollouts/routes.py
+++ b/goosebit/api/v1/rollouts/routes.py
@@ -16,7 +16,8 @@ router = APIRouter(prefix="/rollouts", tags=["rollouts"])
     dependencies=[Security(validate_user_permissions, scopes=["rollout.read"])],
 )
 async def rollouts_get(_: Request) -> RolloutsResponse:
-    return await RolloutsResponse.convert(await Rollout.all().prefetch_related("software"))
+    rollouts = await Rollout.all().prefetch_related("software", "software__compatibility")
+    return RolloutsResponse(rollouts=rollouts)
 
 
 @router.post(
--- a/goosebit/api/v1/software/routes.py
+++ b/goosebit/api/v1/software/routes.py
@@ -1,3 +1,5 @@
+from __future__ import annotations
+
 import random
 import string
 
@@ -22,7 +24,8 @@ router = APIRouter(prefix="/software", tags=["software"])
     dependencies=[Security(validate_user_permissions, scopes=["software.read"])],
 )
 async def software_get(_: Request) -> SoftwareResponse:
-    return await SoftwareResponse.convert(await Software.all().prefetch_related("compatibility"))
+    software = await Software.all().prefetch_related("compatibility")
+    return SoftwareResponse(software=software)
 
 
 @router.delete(
--- a/goosebit/db/models.py
+++ b/goosebit/db/models.py
@@ -1,3 +1,5 @@
+from __future__ import annotations
+
 from enum import IntEnum
 from typing import Self
 from urllib.parse import unquote, urlparse
--- a/goosebit/ui/bff/rollouts/responses.py
+++ b/goosebit/ui/bff/rollouts/responses.py
@@ -1,5 +1,3 @@
-import asyncio
-
 from fastapi.requests import Request
 from pydantic import BaseModel, Field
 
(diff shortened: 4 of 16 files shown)


goosebit/db/models.py
```diff
@@ -130,18 +129,3 @@ async def latest(cls, device: Device) -> Self | None:
             key=lambda x: semver.Version.parse(x.version),
             reverse=True,
         )[0]
-
```
tsagadar, line 133: Is it a good idea to move such logic into the Pydantic schema? The way I saw the schemas until now was that they represent our DTO (data transfer models). As such, they only care about serialisation/deserialisation and validation.

Depending on our namespaces (v1 vs bff) we will also have different schemas/DTOs for the same model (like Software in this case).

b-rowan: The goal here would be to allow a plugin to extend that schema for it's own purposes, and still be able to conform to the internal stuff.  EG for a plugin we want to build, we would need to be able to push alternate one-off software updates to a device.  This means that passing around the `Software` model internally isn't ideal, since creating a new DB item for each devices' one-off updates will clutter up the UI, mess up the update handler, etc.

Maybe in this case there is some way to build a "one-shot" updater, but I still think generalizing this as much as possible would be nice.

tsagadar:
Took my third attempt to wrap my head around this PR - would limited success. I think there are three things in it
1. Simplify the Pydantic model instantiation relying on from_attributes. As an independent commit I think it would be an obvious valuable refactoring.
2. Move logic from the DB models (entities) to the schema (DTO). Not really sure if this is a good idea. Usually logic would remain in the entities.
3. Preparation work for implementing the "initial update for device without unique identifier" use case through a plugin mechanism. It could help to describe this as a feature request - even if the idea is to implement it as a plug-in. This could allow to understand how this can fit into gooseBit. Maybe the feature should even be part of the regular code base and that can be activated through a configuration?

b-rowan:
>     3. Preparation work for implementing the "initial update for device without unique identifier" use case through a plugin mechanism. It could help to describe this as a feature request - even if the idea is to implement it as a plug-in. This could allow to understand how this can fit into gooseBit. Maybe the feature should even be part of the regular code base and that can be activated through a configuration?

It's not specifically that, I actually had a cloudflare integration in mind.  Right now with our internal version (which is very outdated), we have a cloudflare integration which creates tunnels to devices automatically based on UUIDs.  This requires configuring `cloudflared` on the device with a specific private key, and in order to do that we are generating a custom SWU that only writes to that config file, then pushing it to the device.  So the idea would be to check for gooseBit updates for a device, then check for custom SWU file from plugin, and to do that we need a standard internal interface that doesn't require creating a DB entry (as I would like to delete these SWU's off disk ASAP, we can always regenerate them later).

goosebit/db/models.py
b-rowan: I have kept it in the original `Software` model as well now, but never hurts to include it here as well.

b-rowan:
> Simplify the Pydantic model instantiation relying on from_attributes. As an independent commit I think it would be an obvious valuable refactoring.

I think you're right about this for now.  I have made this PR a bit more atomic by removing 2 and 3.  I would really like to be able to get a plugin interface of some kind included here, I want to have our internal stuff up to date, but it seems every foothold I've tried to gain on that side of things causes one issue or another.  If you have any ideas on that side of things LMK.

goosebit/ui/bff/rollouts/responses.py
```diff
@@ -32,6 +30,6 @@ async def convert(cls, request: Request, query, search_filter, total_records):
 
         filtered_records = await query.count()
         rollouts = await query.offset(start).limit(length).all()
-        data = await asyncio.gather(*[RolloutSchema.convert(r) for r in rollouts])
+        data = [RolloutSchema.model_validate(r) for r in rollouts]
```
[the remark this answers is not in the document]
b-rowan: Should be fixed now.  Had to update the pre-fetch in routes.

goosebit/api/v1/software/routes.py
```diff
@@ -22,7 +24,7 @@
     dependencies=[Security(validate_user_permissions, scopes=["software.read"])],
 )
 async def software_get(_: Request) -> SoftwareResponse:
-    return await SoftwareResponse.convert(await Software.all().prefetch_related("compatibility"))
+    return SoftwareResponse(software=await Software.all().prefetch_related("compatibility"))
```
tsagadar, line 27: Just cosmetics: in the rest of the code this is two lines first assigning using a variable software.

pyproject.toml
```diff
@@ -37,7 +37,6 @@ pre-commit = "^3.6.2"
 flake8 = "7.1.0"
 mypy = "^1.11.2"
 types-pyyaml = "^6.0.12.20240808"
-types-aiofiles = "^24.1.0.20240626"
```
tsagadar, line 40: Could be a separate commit?

goosebit/api/v1/rollouts/routes.py
```diff
@@ -16,14 +16,18 @@
...
+    rollouts = await Rollout.all().prefetch_related("software", "software__compatibility")
+    return RolloutsResponse(rollouts=rollouts)
 
 
 @router.post(
     "",
     dependencies=[Security(validate_user_permissions, scopes=["rollout.write"])],
 )
 async def rollouts_put(_: Request, rollout: RolloutsPutRequest) -> RolloutsPutResponse:
+    software = await Software.filter(id=rollout.software_id)
```
tsagadar, line 28: Could be a separate commit?

tsagadar (approved):
Looks good. Just minor issues that are optional to address in my opinion.

tsagadar:
> > Simplify the Pydantic model instantiation relying on from_attributes. As an independent commit I think it would be an obvious valuable refactoring.
> 
> I think you're right about this for now. I have made this PR a bit more atomic by removing 2 and 3. I would really like to be able to get a plugin interface of some kind included here, I want to have our internal stuff up to date, but it seems every foothold I've tried to gain on that side of things causes one issue or another. If you have any ideas on that side of things LMK.

@b-rowan let's discuss this through another channel. Can you send me an email to the address from my commits?

b-rowan:
> Can you send me an email to the address from my commits?

Done.  Sent from my upstreamdata.ca email.

goosebit/api/v1/software/routes.py
```diff
@@ -1,3 +1,5 @@
+from __future__ import annotations
```
easybe, line 1: Why are these lines needed?

b-rowan: Not really sure why I've been adding them, just a good failsafe.  Need to modify my IDE settings to stop complaining about them I think.

PR #132, opened 2024-09-10
Add support for RAUC file parsing

### LINKED ISSUE #131: Does goosbit support RAUC packages?
Hello, 
We are currently using Hawkbit for our machine updates with RAUC packages. We think of migrating to goosebit and use it in a production environment but from our first staging trials we noticed that it only supports SWUpdates packages.
Is it possible to provide another format for updates?

b-rowan:
Allow parsing of RAUC files using `PySquashFsImage`. 

Closes #131


goosebit/updates/swdesc/rauc.py
```diff
@@ -0,0 +1,42 @@
...
+        swdesc_attrs["hash"] = await sha1_hash_file(f)
+    return swdesc_attrs
+
+
+def parse_descriptor(manifest: configparser.ConfigParser):
+    swdesc_attrs = {}
+    try:
+        # specified as optional in the RAUC docs
+        swdesc_attrs["version"] = semver.Version.parse(manifest["update"].get("version") + ".0")
+        swdesc_attrs["compatibility"] = [{"hw_model": "default", "hw_revision": manifest["update"]["compatible"]}]
```
tsagadar, line 37: Having a hard-coded hw_model "default" means that only a single hardware (platform) is supported to be managed in goosebit. Seems very limiting.

goosebit/updates/swdesc/rauc.py
```diff
@@ -0,0 +1,42 @@
...
+from PySquashfsImage import SquashFsImage
+
+from .func import sha1_hash_file
+
+MAGIC = b"hsqs"
+
+logger = logging.getLogger(__name__)
+
+
+async def parse_file(file: Path):
```
tsagadar, line 15: There should be at least one test for this parsing code (as there is for swu images)

goosebit/updates/__init__.py
```diff
@@ -20,10 +20,10 @@ async def create_software_update(uri: str, temp_file: Path | None) -> Software:
 
     # parse swu header into update_info
     if parsed_uri.scheme == "file" and temp_file is not None:
-        try:
-            update_info = await swdesc.parse_file(temp_file)
-        except Exception:
-            raise HTTPException(422, "Software swu header cannot be parsed")
+        # try:
```
tsagadar, line 23: Should not merge commented out code

goosebit/updates/swdesc/__init__.py
```diff
@@ -0,0 +1,33 @@
...
+        async with aiofiles.tempfile.NamedTemporaryFile("w+b") as f:
+            await f.write(file.content)
+            return await parse_file(Path(str(f.name)))
+
+
+async def parse_file(file: Path):
+    async with await open_file(file, "r+b") as f:
+        magic = await f.read(4)
+    if magic == swu.MAGIC:
+        image_format = "SWU"
```
tsagadar, line 24: Why use string literals and not the `SoftwareImageFormat` enum?

tsagadar (changes requested):
Not sure if it is a bit too aggressive to already add RAUC support. It is not only about the file format but also about being able to document / test its usage. I'd expect minor differences in comparison with SWUpdate.

Some more open things
- Adding db migrations (as discussed in #133)
- Adjusting documentation
- Exposing image type in UI? Currently the type is stored but never used
- Squash the "[pre-commit.ci]" commit? Or is the idea not to run the code formatting tools and have a lot of those commits on main?

goosebit/updates/swdesc/rauc.py
b-rowan: Added one, it doesnt do the compatibility parsing, see https://github.com/UpstreamDataInc/goosebit/pull/132#discussion_r1754789482

easybe:
I think it is a good idea to support RAUC as long as we don’t have to completely refactor the application at this point. IMHO we should focus on making what we have production ready and then go from there.

b-rowan:
> I think it is a good idea to support RAUC as long as we don’t have to completely refactor the application at this point. IMHO we should focus on making what we have production-ready and then go from there.

I agree here.  This was relatively easy to throw together, so if we don't want to add this right now I will leave as a draft for reference.

tsagadar:
Another thing I realized today: goosebit does log message parsing (to detect download progress / update status) - this also would require adjustments. However, the parsing ideally gets reduced to only extract the progress.

Tristoris:
> > Hello! I have read through the changes and issues and also used today's master version and cherry-picked the two commits and rauc seemed to work with goosebit (I did have to resolve a few conflicts though). Are there any issues with this pull request so that the changes are not on master yet? Is something missing to apply the changes/ has something else to be done before closing this?
> 
> Not entirely sure. Our main goal was to ensure that gooseBit as a whole is production ready before we attempt to add complexity with something like this. If you are able to provide log examples that I can work on parsing with this we likely could add it, but I think there also needs to be consideration made as to how we track which files are which types, and which devices want what type of updates.

I understand. However isn't it the job of the update manager to know what files are supported by what devices? Since goosebit is really just a way to bring that file to the device / decide what devices will receive what updates. The only time it has to do something with the file type is when it tries to read the header for information about the uploaded update, nothing more; which was accounted for in those two commits here.

b-rowan:
> However isn't it the job of the update manager to know what files are supported by what devices?

I think partially, yes.  I worry about edge cases, like what if a user has 2 devices, one which supports RAUC and one which supports SWUpdate, and the user uploads 2 files with the same versioning?  It's a good idea to be at least partially aware of this information, if we can save the user from having the device download invalid updates I think that is a good idea.

Curious what others think here though, I have no opposition to merging this as-is, but we should deliberate.

easybe:
I think it is fine to push this forward. I have started to whip up da device docker image for RAUC, so we could even have E2E tests.

goosebit/updates/swdesc/rauc.py
easybe: My suggestion is to support the convention `compatible=<hw_model>-<hw_revision>`

easybe:
The RAUC docs use `version=2016.08-1` as an example. Perhaps we need to add support for that format as well.

easybe:
> This all looks good to me, we can add log parsing at a later time. I won't be able to do anything on this until next week, as I am away this week, but this seems good to merge.

OK, I will finish the PR then.

As mentioned [here](https://github.com/UpstreamDataInc/goosebit/pull/132#issuecomment-3252428965), log parsing seems to work just fine.

tests/updates/swu/test_swdesc.py
```diff
@@ -0,0 +1,14 @@
...
+from goosebit.updates.swdesc.rauc import parse_file
+
+
+@pytest.mark.asyncio
+async def test_parse_software_header():
+    resolved = await Path(__file__).resolve()
+    swdesc_attrs = await parse_file(resolved.parent / "software-header.raucb")
+    assert str(swdesc_attrs["version"]) == "8.8.1-11-g8c926e5+188370"
+    assert swdesc_attrs["compatibility"] == [
+        {"hw_model": "default", "hw_revision": "rauc-test-goosebit"},
```
easybe, line 13: Adapt

UpstreamData (dismissed):

UpstreamData (approved):

easybe:
@Tristoris, with the latest release we now have RAUC support. Let us know if that works for you.

Tristoris:
image

Hi! I wanted to ask whether it is intended that the compatibility text is split like that "r - aspberrypi4" (My set compatibility was raspberrypi4 and raspberrypi4 during build). I also noticed that the devices are not getting the updates i uploaded, which could be explained by the compatibility section there.

PR #135, opened 2024-09-13
Remove aiofiles dependency

b-rowan:
Remove dependency on aiofiles, `NamedTemporaryFile` was causing issues. Replace temporary files with a temporary file in `artifacts/tmp`.

### DIFF
--- a/goosebit/api/v1/software/routes.py
+++ b/goosebit/api/v1/software/routes.py
@@ -1,5 +1,7 @@
-import aiofiles
-from anyio import Path
+import random
+import string
+
+from anyio import Path, open_file
 from fastapi import APIRouter, File, Form, HTTPException, Security, UploadFile
 from fastapi.requests import Request
 
@@ -69,11 +71,16 @@ async def post_update(_: Request, file: UploadFile | None = File(None), url: str
         # local file
         artifacts_dir = Path(config.artifacts_dir)
         file_path = artifacts_dir.joinpath(file.filename)
-
-        async with aiofiles.tempfile.NamedTemporaryFile("w+b") as f:
-            await f.write(await file.read())
-            absolute = await file_path.absolute()
-            software = await create_software_update(absolute.as_uri(), Path(f.name))
+        tmp_file_path = artifacts_dir.joinpath("tmp", ("".join(random.choices(string.ascii_lowercase, k=12)) + ".tmp"))
+        await tmp_file_path.parent.mkdir(parents=True, exist_ok=True)
+        file_absolute_path = await file_path.absolute()
+        tmp_file_absolute_path = await tmp_file_path.absolute()
+        try:
+            async with await open_file(tmp_file_path, "w+b") as f:
+                await f.write(await file.read())
+            software = await create_software_update(file_absolute_path.as_uri(), tmp_file_absolute_path)
+        finally:
+            await tmp_file_path.unlink(missing_ok=True)
     else:
         raise HTTPException(422)
 
--- a/goosebit/updates/__init__.py
+++ b/goosebit/updates/__init__.py
@@ -19,7 +19,9 @@ async def create_software_update(uri: str, temp_file: Path | None) -> Software:
     parsed_uri = urlparse(uri)
 
     # parse swu header into update_info
-    if parsed_uri.scheme == "file" and temp_file is not None:
+    if parsed_uri.scheme == "file":
+        if temp_file is None:
+            raise HTTPException(500, "Temporary file missing, cannot parse file information")
         try:
             update_info = await swdesc.parse_file(temp_file)
         except Exception:
@@ -30,7 +32,6 @@ async def create_software_update(uri: str, temp_file: Path | None) -> Software:
             update_info = await swdesc.parse_remote(uri)
         except Exception:
             raise HTTPException(422, "Software swu header cannot be parsed")
-
     else:
         raise HTTPException(422, "Software URI protocol unknown")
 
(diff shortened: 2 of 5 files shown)


goosebit/api/v1/software/routes.py
```diff
@@ -69,11 +71,15 @@ async def post_update(_: Request, file: UploadFile | None = File(None), url: str
...
 
-        async with aiofiles.tempfile.NamedTemporaryFile("w+b") as f:
+        async with await open_file(tmp_file_path, "w+b") as f:
             await f.write(await file.read())
-            absolute = await file_path.absolute()
-            software = await create_software_update(absolute.as_uri(), Path(f.name))
+        absolute = await file_path.absolute()
+        tmp_absolute = await tmp_file_path.absolute()
+        software = await create_software_update(absolute.as_uri(), tmp_absolute)
+        await tmp_file_path.unlink(missing_ok=True)
```
tsagadar, line 82: Move temp file deletion in a finally block to ensure that it always gets deleted.

goosebit/updates/swdesc.py
```diff
@@ -68,9 +71,14 @@ async def parse_file(file: Path):
...
         file = await c.get(url)
-        async with aiofiles.tempfile.NamedTemporaryFile("w+b") as f:
+        artifacts_dir = Path(config.artifacts_dir)
+        tmp_file_path = artifacts_dir.joinpath("tmp", ("".join(random.choices(string.ascii_lowercase, k=12)) + ".tmp"))
+        await tmp_file_path.parent.mkdir(parents=True, exist_ok=True)
+        async with await open_file(tmp_file_path, "w+b") as f:
             await f.write(file.content)
-            return await parse_file(Path(str(f.name)))
+        file_data = await parse_file(Path(str(f.name)))
+        await tmp_file_path.unlink()
```
tsagadar, line 80: Move temp file deletion in a finally block to ensure that it always gets deleted.

goosebit/updates/__init__.py
```diff
@@ -19,7 +19,9 @@ async def create_software_update(uri: str, temp_file: Path | None) -> Software:
     parsed_uri = urlparse(uri)
 
     # parse swu header into update_info
-    if parsed_uri.scheme == "file" and temp_file is not None:
+    if parsed_uri.scheme == "file":
+        if temp_file is None:
+            raise HTTPException(422, "Temporary file missing, cannot parse file information")
```
tsagadar, line 24: I think a better error code would be 500. This is an internal server issue - nothing caused by the client request.

tsagadar (changes requested):

easybe:
> Remove dependency on aiofiles, `NamedTemporaryFile` was causing issues. Replace temporary files with a temporary file in `artifacts/tmp`.

What kind of issues?

goosebit/api/v1/software/routes.py
```diff
@@ -69,11 +71,15 @@ async def post_update(_: Request, file: UploadFile | None = File(None), url: str
...
         file_path = artifacts_dir.joinpath(file.filename)
+        tmp_file_path = artifacts_dir.joinpath("tmp", ("".join(random.choices(string.ascii_lowercase, k=12)) + ".tmp"))
+        await tmp_file_path.parent.mkdir(parents=True, exist_ok=True)
 
-        async with aiofiles.tempfile.NamedTemporaryFile("w+b") as f:
+        async with await open_file(tmp_file_path, "w+b") as f:
             await f.write(await file.read())
-            absolute = await file_path.absolute()
-            software = await create_software_update(absolute.as_uri(), Path(f.name))
+        absolute = await file_path.absolute()
```
easybe, line 79: I don’t find `absolute` a good variable name. It doesn’t suggest what it is storing. Absolut Vodka? 😉

b-rowan:
> > Remove dependency on aiofiles, `NamedTemporaryFile` was causing issues. Replace temporary files with a temporary file in `artifacts/tmp`.
> 
> 
> 
> What kind of issues?

Was running into permissions errors.  It wasn't entirely aiofiles, but the way we were opening and closing files.  We opened the temp file at the top level, then tried to open it again in a function called inside that with block, and that was causing issues.

In this case though, I don't really see any reason to use aiofiles or temp files here, there doesn't seem like much benefit...

goosebit/api/v1/software/routes.py
b-rowan: That would be a heck of a variable 😂

I'll get this renamed.

tsagadar:
LGTM. Great to have that dependency removed!

tsagadar (approved):

goosebit/api/v1/software/routes.py
```diff
@@ -69,11 +71,16 @@ async def post_update(_: Request, file: UploadFile | None = File(None), url: str
         # local file
         artifacts_dir = Path(config.artifacts_dir)
         file_path = artifacts_dir.joinpath(file.filename)
-
-        async with aiofiles.tempfile.NamedTemporaryFile("w+b") as f:
-            await f.write(await file.read())
-            absolute = await file_path.absolute()
-            software = await create_software_update(absolute.as_uri(), Path(f.name))
+        tmp_file_path = artifacts_dir.joinpath("tmp", ("".join(random.choices(string.ascii_lowercase, k=12)) + ".tmp"))
```
easybe, line 74: This change makes the artifact directory mandatory. Before, it was optional (when only using external artifacts) as temporary files were created in `/tmp` :disappointed:

b-rowan: Didn't even think about this.  Maybe I can generate a temporary file location other than this?

easybe: IMHO it is best to use a standard temp location. But, we should make sure it works on all OSes.

PR #139, opened 2024-09-22
Fix version parsing in SWU

sbabic:
The parser raises an error if version is not set according to the semantic version rules. However, this is not mandatory in SWUpdate and any version string is allowed. Set version from sw-description if a semantic version cannot be parsed.

### DIFF
--- a/goosebit/updates/__init__.py
+++ b/goosebit/updates/__init__.py
@@ -1,8 +1,10 @@
 from __future__ import annotations
 
+import logging
 from urllib.parse import unquote, urlparse
 from urllib.request import url2pathname
 
+import semver
 from anyio import Path
 from fastapi import HTTPException
 from fastapi.requests import Request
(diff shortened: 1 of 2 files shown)


b-rowan:
I think supporting arbitrary versions is a good idea, but the problem arises when trying to decide on the "latest" version for a device when using that update mode.  The `Software.latest()' function will need to be updated to handle the arbitrary version case somehow.

https://github.com/UpstreamDataInc/goosebit/blob/master/goosebit/db/models.py#L125-L134

sbabic:
I see - but setting a rule for arbitrary version string could be a challenge. An alternative could be to relax the version and convert a numbered version into a semantic version. SWUpdate near semantic version supports for comparison "legacy" versions, and this is a version in the format of <major>.<minor>.<patch>.<build>. Each field (apart major) is optional, so a version "1" or "1.0" are allowed for version comparison. So in case the parser of semantic version (for example, "0.1") raises an error, we could try to convert it (setting 0 for not set field, that is "0.1" becomes "0.1.0") so that could be accepted again by the parser and the rest of logic works.

b-rowan:
Yeah, I think that makes sense.

In that case, we should take a look at `awesomeversion`, seems like it can handle this case.

https://pypi.org/project/awesomeversion/

tsagadar:
Superseded by #145

PR #141, opened 2024-09-23
Handle DataTables query parameters with `Depends`

b-rowan:
Use `Pydantic` and `Depends` to automatically parse the DataTables AJAX request to a typed format.

Precursor to a fix required in #138, which should use query parameters to validate software compatibility.


tsagadar (approved):
LGTM

PR #145, opened 2024-09-29
Use more flexible version parsing

tsagadar:
SWUpdate does not restrict versions to follow semver.org. So using a more flexible version parsing by making minor and patch elements optional.

Addresses same issue as #139


b-rowan (approved):
Nice, I like this for now.

Issue #146, opened 2024-09-29
Device's "Target Software" only set when directly assigning software


easybe:
I guess it should also show the software version that shall be installed when "rollout" or "latest" is selected.
Or, alternatively rename the column?

b-rowan:
I think in the future this will be a computed field that references either a software version directly, or a rollout software version.

tsagadar:
Had a quick look at the code. This is a regression of commit 1fb6368 in `DeviceSchema`. Target software before was looked up like this:
```
        manager = await get_update_manager(device.uuid)
        _, target_software = await manager.get_update()
```

@b-rowan not sure if using `validate_model` was a good choice in generating responses (the function name also suggests that it is rather used for deserializing requests). It reduces the flexibility and sometimes requires to load more than is being serialized.

PR #148, opened 2024-09-30
Select target firmware based on device state

b-rowan:
Fixes: #146


easybe:
The `#` is missing in the commit message.

tsagadar (approved):
LGTM. I hope to find the time soon to increase test coverage (also including the /bff parts) so that for such changes we also could easily add an additional test to prevent future regressions.

Issue #137, opened 2024-09-13
Online status ("Up") flickering 


easybe:
At least with our deployment, even though my device is obviously communicating with the server (as indicated by "Last seen") the "Up" status turns red from time to time.

b-rowan:
The `up` value just  checks if the device has communicated within whatever value is set for poll time on that device.  We could change it to  check against default poll time, plus a couple of seconds.

easybe:
I think there should be quite a margin, maybe in percent? Eventually, we plan to use a poll time of 1 day or more, so we should keep that in mind.

b-rowan:
I think anywhere from 1.2 - 1.5X the poll time would be fine.  We could also make that configurable pretty easily, just add it to the pydantic settings, something like `poll_time_online_multiplier`?

easybe:
I don't think it should be configurable. It should "just work" for short and long intervals.

PR #149, opened 2024-09-30
Add an extra 10 seconds to online check

b-rowan:
Fixes: #137


easybe:
When starting a software update, the dot still seems to turn red:
up_status

tsagadar:
Changes included in #137

Issue #147, opened 2024-09-29
"Home" tab redundant?


easybe:
At least currently, the content of the "Home" tab is almost identical to that of "Devices". Remove it (for now)?

b-rowan:
No opposition here to removing it for now, but I think generally we should consider what the most important aspects of that page could be.  What are the things that are important to a user at first glance, things like device state, rollout status, etc.  I think eventually we will have to move that page away from the "single table" design that we use currently, and over to something more akin to a summary page.

PR #150, opened 2024-10-02
Remove home tab

tsagadar:
Currently, there is no additional value in the home tab. Let's bring it back once there are specific ideas what use cases it is for.

Issue: #147


easybe:
Thanks. I think we are getting closer to an MVP!

PR #151, opened 2024-10-02
Remove UnknownUpdateManager

### LINKED ISSUE #12: Use of UnknownUpdateManager
What is the idea behind `UnknownUpdateManager`? Looks like it only would get used if a device reports with the controllerId `unknown`.

Can I remove this (and its base class)?

tsagadar:
This use case is neither tested nor documented. So better remove it for the time being.

Resolves #12


b-rowan (approved):
For now looks good, but I want to revisit this in the future, either as a plugin (`goosebit-device-initializer` or something), or as a properly supported feature.

Issue #153, opened 2024-10-03
DataTables warnings


easybe:
![Screenshot from 2024-10-03 09-21-47](https://github.com/user-attachments/assets/843adf18-19f6-4e03-b2c7-7cd323ce06af)

I get these annoying popups when e.g. the user session is no longer valid because of a restart of the dev server. But, I  have also seen them after having the browser in the background.

I think we should suppress them and maybe add some indicator making the user aware of communication problems with the backend.

b-rowan:
Yeah, for sure.  I think we should try refreshing the page when this happens maybe?  Most of the time this happens it is because of an invalid session...

PR #155, opened 2024-10-03
Reload page when datatables show auth error from BFF

b-rowan:
Fixes: #153

### DIFF
--- a/goosebit/ui/templates/nav.html.jinja
+++ b/goosebit/ui/templates/nav.html.jinja
@@ -60,7 +60,18 @@
                 box-shadow: none!important;
             }
         </style>
-        <script>const TABLE_UPDATE_TIME = 3000;</script>
+        <script>
+            const TABLE_UPDATE_TIME = 3000;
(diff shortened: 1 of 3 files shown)


goosebit/ui/templates/nav.html.jinja
```diff
@@ -60,7 +60,18 @@
                 box-shadow: none!important;
             }
         </style>
-        <script>const TABLE_UPDATE_TIME = 3000;</script>
+        <script>
+            const TABLE_UPDATE_TIME = 3000;
+            DataTable.ext.errMode = function ( e, settings, helpPage, message ) {
+                if (e.jqXHR.status == 401) {
+                    window.location.reload();
+                } else if (e.jqXHR.status >= 200 & e.jqXHR.status < 300 ) {
```
easybe, line 68:
```suggestion
                } else if (e.jqXHR.status >= 200 && e.jqXHR.status < 300 ) {
```

easybe (changes requested):
Looks good otherwise

easybe (approved):

PR #156, opened 2024-10-05
Fix sorting software by version

tsagadar:
Opted to use server-side in-memory sorting relying on the semver package to be consistent with the "latest" update mode.

Also added additional tests, fixed a few issues, and limited the supported query params in the /ui/bff endpoints.

### DIFF
--- a/goosebit/ui/bff/rollouts/responses.py
+++ b/goosebit/ui/bff/rollouts/responses.py
@@ -19,11 +19,15 @@ class BFFRolloutsResponse(BaseModel):
         if dt_query.search.value:
             query = query.filter(search_filter(dt_query.search.value))
 
+        filtered_records = await query.count()
+
         if dt_query.order_query:
             query = query.order_by(dt_query.order_query)
 
-        filtered_records = await query.count()
-        rollouts = await query.offset(dt_query.start).limit(dt_query.length).all()
+        if dt_query.length is not None:
+            query = query.limit(dt_query.length)
+
+        rollouts = await query.offset(dt_query.start).all()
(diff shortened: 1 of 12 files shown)


goosebit/ui/bff/rollouts/responses.py
```diff
@@ -22,8 +22,10 @@ async def convert(cls, dt_query: DataTableRequest, query: QuerySet, search_filte
...
-        filtered_records = await query.count()
-        rollouts = await query.offset(dt_query.start).limit(dt_query.length).all()
+        if dt_query.length is not None:
+            query = query.limit(dt_query.length)
+
+        rollouts = await query.offset(dt_query.start).all()
         data = [RolloutSchema.model_validate(r) for r in rollouts]
 
-        return cls(data=data, draw=dt_query.draw, records_total=total_records, records_filtered=filtered_records)
+        return cls(data=data, draw=dt_query.draw, records_total=total_records, records_filtered=len(rollouts))
```
b-rowan, line 31: I think we should confirm what value this actually updates, my understanding was that this was for pagination, so `Showing 10 of {records_filtered} records, filtered from {records_total} total records`.  I would assume that datatables would be able to check the length of an array.

b-rowan (approved):
Good, but want to confirm datatables functionality.

goosebit/ui/bff/rollouts/responses.py
tsagadar: Correct. Fixed this issue. Also tested it with generated sample data. Will add the capability to generate sample data through additional PR.

PR #162, opened 2024-10-10
Add Azure/EntraID route for SAML

b-rowan:
Adds support for SAML via the optional `azure` extra. This can be configured with the `config.auth.azure` options.


rettichschnidi:
Wow, that was fast. Great, thanks a lot! However, I am not (yet) in a position to actually test this. We (@easybe, @tsagadar , worst case... I) will need to talk to our IT department first.

b-rowan:
It's actually not even ready yet, it will need some fixing, I think the requests I'm trying from the `msal` lib are probably not right, I will need someone with an actual AzureAD account to try them out and add the proper handling.  For now this is just a button, some settings, some checks, and dependency handling.

Issue #164, opened 2024-10-18
Remove "Last IP" column


easybe:
Due to GDPR-related concerns, we would like to be able to disable the "Last IP" column or get rid of it altogether.

b-rowan:
Sure.  It is useful to us internally because we have to be able to directly access these devices sometimes to help our clients, but we are also in a very different scenario.  This is probably a good option for a plugin at some point.

rettichschnidi:
Maybe something in the spirit of https://github.com/DigitaleGesellschaft/Anonip could address the concerns sufficiently?

easybe:
> Maybe something in the spirit of https://github.com/DigitaleGesellschaft/Anonip could address the concerns sufficiently?

The thing is, in our setup, the IP addresses are not resolved correctly, so we came to the conclusion that removing the column altogether would be the easiest for us 🤷‍♂️

rettichschnidi:
> The thing is, in our setup, the IP addresses are not resolved correctly, ...

Thanks to the (AWS?) reverse proxy I guess? Maybe it will change when IPv6 gets used?

easybe:
> > The thing is, in our setup, the IP addresses are not resolved correctly, ...
> 
> Due to the reverse proxy I guess?
> 
> As you said we'd rather just remove the storage (or display only?) of the IP addresses entirely, but if anyone else has this problem: Using/Supporting [X-Forwarded-For](https://en.wikipedia.org/wiki/X-Forwarded-For) would resolve this issue.

Yes, we documented that here: https://github.com/UpstreamDataInc/goosebit/blob/master/docker/demo/nginx/site.conf

IPv6 is indeed a problem.

Anyway, with AWS such things are not always that straightforward. But, we can have a look offline if you are interested.

b-rowan:
So the course of action we want to take is removing it, correct?  I have no opposition to that in the main branch, so I can throw together a quick PR to remove it.

Additionally, X-Forwarded-For is something I believe the internal version we were using supported.  The nice part of doing this as a plugin is that it's basically just a middleware, so it is really easy to inject.

jameshilliard:
> So the course of action we want to take is removing it, correct?

I'm not a big fan of removing this, a config flag I think is preferable as this is likely to be useful debugging info for a lot of deployments.

> The nice part of doing this as a plugin is that it's basically just a middleware, so it is really easy to inject.

The bigger issue I think is the data storage, a plugin would need to use a separate table probably.

[...]

PR #165, opened 2024-10-19
Remove IP tracking

b-rowan:
Fixes: #164

### DIFF
--- a/goosebit/db/migrations/models/1_20241021082844_drop_last_ip.py
+++ b/goosebit/db/migrations/models/1_20241021082844_drop_last_ip.py
@@ -1,0 +1,13 @@
+from tortoise import BaseDBAsyncClient
+
+
+async def upgrade(db: BaseDBAsyncClient) -> str:
+    return """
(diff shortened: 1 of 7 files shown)


easybe (changes requested):
Please use a meaningful name for the migration, e.g. `1_20241019081940_drop_last_ip.py`

LGTM otherwise

jameshilliard:
I'm thinking it's better to make this a config flag to disable rather than remove the columns entirely.

easybe (approved):

easybe:
> I'm thinking it's better to make this a config flag to disable rather than remove the columns entirely.

Both is fine for us. However, I guess this can easily be brought back when needed. Anyway, IMHO we should focus on getting the core functionality as stable/usable as possible, so less is more. For us, the most important thing is scalability, as we have a lot of devices.

b-rowan:
I think based on recent discussion that this should be reworked to a config option.  Should be pretty simple, it will only instantiate the middleware if the config flag is enabled.

PR #167, opened 2024-10-26
Fix online status ("Up") flickering bug

tsagadar:
Fixes: #137

PR does two things
1. Includes adding a 10 seconds safety margin from PR #149. 
2. Prevents poll time to be adjusted when assigning software to device

These two things together should get rid of the flickering.

Note: the log file handling likely needs similar changes. Will create a new ticket for the log files as it could require some discussions first.


b-rowan (approved):

easybe:
@tsagadar deployed on our instance.

PR #170, opened 2024-10-28
Allow configuring ip tracking by loading columns from bff API

### LINKED ISSUE #164: Remove "Last IP" column

b-rowan:
IP tracking can now be disabled with the `track_device_ip` config flag. The UI also has to hide this column, as it will no longer show up in data. To handle this, the table columns are dynamically created with a request to the bff API.

Fixes #164

### DIFF
--- a/goosebit.yaml
+++ b/goosebit.yaml
@@ -16,6 +16,9 @@ poll_time_default: 00:01:00
 # and device logs get updated.
 poll_time_updating: 00:00:05
 
+# Whether to track the IP of the device when it polls.  Useful for debugging, but can be turned off for privacy.
+# track_device_ip: true
+
 # Secret key used for parsing user sessions. It is HIGHLY advised to pass this as an environment variable instead.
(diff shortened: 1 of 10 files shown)


b-rowan:
I would like to extend this style of creating tables dynamically from the BFF API, as it allows us to extend them in the future, including allowing users to configure what columns should be shown, which can reduce clutter.

easybe (changes requested):
I tried it (on top of current master), and, "Online/Last seen" no longer seems to work.

easybe (approved):
Works again, thanks!

easybe:
> I would like to extend this style of creating tables dynamically from the BFF API, as it allows us to extend them in the future, including allowing users to configure what columns should be shown, which can reduce clutter.
> 
> Let me know thoughts on this. If we want to go in this direction I will do this for the other pages as well.

I don't have an opinion on this. But, maybe @tsagadar has one.

PR #173, opened 2024-10-29
Fix log handling issues with websocket auth and connection closing

### LINKED ISSUE #168: Log file handling
My assumption how the log file feature should work:
1. Incrementally displays the SWUpdate logs during an active update process
2. Displays the SWUpdate logs of the last update if no update process is active

Currently number 2. would be broken. Before looking into fixing this I wanted to make sure that we expect the same behavior.

b-rowan:
Websocket authentication wasn't handling oauth properly because of a positional `Request`. Log subscribers may also break when connections are closed when the remove call is combined, needs to be separate.

Fixes #168

### DIFF
--- a/goosebit/auth/__init__.py
+++ b/goosebit/auth/__init__.py
@@ -16,7 +16,11 @@ from goosebit.settings.schema import User
 logger = logging.getLogger(__name__)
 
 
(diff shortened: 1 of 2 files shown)


tsagadar (approved):
Great! Checked and it works. Only started to scratch the surface to fix the issue. So no need to stay up late ;-)

PR #177, opened 2024-11-09
Log handling rework

tsagadar:
Goals of this PR

- Reduce dependency on SWUpdate-specific log parsing
- Simplify code in the space of log handling
- Introduce automated tests to check log handling. Ideally the test would rely on the WebSocket endpoint to do the assertions but failed to do so.

Not 100% if commit e1c94c843 could introduce a regression. @b-rowan do you know the background here?

### DIFF
--- a/goosebit/updater/controller/v1/routes.py
+++ b/goosebit/updater/controller/v1/routes.py
@@ -125,33 +125,30 @@ async def deployment_feedback(
             await updater.deployment_action_success()
             await updater.update_device_state(UpdateStateEnum.FINISHED)
 
-            # not guaranteed to be the correct rollout - see next comment.
             rollout = await updater.get_rollout()
             if rollout:
                 if rollout.software == reported_software:
                     rollout.success_count += 1
                     await rollout.save()
                 else:
+                    # edge case where device update mode got changed while update was running
                     logging.warning(
                         f"Updating rollout success stats failed, software={reported_software.id}, device={updater.dev_id}"  # noqa: E501
                     )
 
-            # setting the currently installed version based on the current assigned software / existing rollouts
-            # is problematic. Better to assign custom action_id for each update (rollout id? software id? new id?).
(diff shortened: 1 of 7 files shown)


goosebit/updater/controller/v1/routes.py
```diff
@@ -184,11 +188,11 @@ async def download_artifact(_: Request, updater: UpdateManager = Depends(get_upd
...
     if software is None:
         raise HTTPException(404)
-    if software.local:
-        return FileResponse(
-            software.path,
-            media_type="application/octet-stream",
-            filename=software.path.name,
-        )
-    else:
-        return RedirectResponse(url=software.uri)
```
b-rowan, line 194: I don't think this can be removed, this is for remote file handling, and the polling endpoint always returns this endpoint (it never returns a remote endpoint).

b-rowan (changes requested):
Just need to make sure we still have remote file handling, other than that I think this is fine.

b-rowan:
> Not 100% if commit e1c94c8 could introduce a regression.

I don't think so, but the logs will get some "Skipped Update" spam.  That's ok though.

goosebit/updater/controller/v1/routes.py
tsagadar: The only place we create an URL to this endpoint is here - and this is specific for `software.local`. What am I missing?

```
async def generate_chunk(request: Request, updater: UpdateManager) -> list:
    _, software = await updater.get_update()
    if software is None:
        return []
    if software.local:
        href = str(
            request.url_for(
                "download_artifact",
                dev_id=updater.dev_id,
            )
        )
    else:
        href = software.uri
```

tsagadar:
@b-rowan added three additional small commits to this PR.

goosebit/updater/controller/v1/routes.py
b-rowan: I think you're right.

b-rowan (approved):
LGTM 🚀

Issue #175, opened 2024-11-08
Device Authentication support


fhuebner:
First: We are currently using Hawkbit and are really happy to see an alternative evolving!

In our current setup each device has a dedicated Authentication Token which is mandatory to allow a device to connect to Hawkbit. Is something like this planed to be integrated into goosebit?

b-rowan:
This is a great idea, I know I've seen this in hawkbit docs, but not entirely sure how they implement it.  Reading Java is not my favorite pastime, but I'm sure we can get it done.

Being a user of this, how would you like to see this implemented?  Per device tokens that could be updated or generated via the API?  A global level token?  Is there a format that works best for you?

fhuebner:
We use the per device security-token, which is normally generated at device creation within hawkbit automatically. This can afterwards also be overwritten with a API-call.
I don’t know how good this procedure would fit into the goosebit: “create device on first connection” philosophy?
Maybe a additional API-Endpoint to manually create devices (and setting the security-token) would be needed for this feature?
Onboarding a device that needs a security-token but is only generated on the first appearance is some sort of a egg-chicken problem, right?

easybe:
We ar using mTLS/x.509 certificate authentication on reverse proxy level. Would that be an option for you?

b-rowan:
> We use the per device security-token, which is normally generated at device creation within hawkbit automatically. This can afterwards also be overwritten with a API-call. I don’t know how good this procedure would fit into the goosebit: “create device on first connection” philosophy? Maybe a additional API-Endpoint to manually create devices (and setting the security-token) would be needed for this feature? Onboarding a device that needs a security-token but is only generated on the first appearance is some sort of a egg-chicken problem, right?

Meh, I don't imagine this is a huge deal.  You should be able to create devices via the API (can't remember if we've done this yet, but should be a thing at some point), and in doing so should just be able to set a device auth token.

The other option is that we have a setting for `admins_accept_devices`, in which the devices could configure their own token on initial setup or post to `configData`, but it would need to be approved by an admin user.  The other advantage of this is that you could easily implement key rotations in the future, to prevent the keys being intercepted for too long.

tsagadar:
If we want to extend this, then I'd propose to stick close to the hawkBit approach: https://eclipse.dev/hawkbit/concepts/authentication/

[...]

PR #193, opened 2024-11-12
Add device `TargetToken` auth

b-rowan:
Adds support for the `TargetToken` auth header.

This can be set up in 3 modes (or disabled altogether, which is the default):
`setup` - Setup mode is for setting up or migrating from an existing solution (such as hawkbit).  Setup mode will auto register all devices that poll, and attach the auth token they poll with to their device information in the database.
`lax` - Lax mode will handle auth tokens on devices that have them set, and will error if those devices don't send a valid token, but devices that were set up without a token will not require one.
`strict` - Strict mode requires a token for every device.  All devices will encounter errors if they submit invalid tokens, and new devices are not registered via the DDI endpoints.  All devices must be set up via the `/device` PUT endpoint, and all devices must have authentication tokens.

Fixes: #175

### DIFF
--- a/goosebit/updater/routes.py
+++ b/goosebit/updater/routes.py
@@ -4,7 +4,6 @@ from fastapi import APIRouter, Depends, HTTPException
 from fastapi.requests import Request
 
 from goosebit.device_manager import DeviceManager, get_device
-from goosebit.settings import config
 from goosebit.settings.schema import DeviceAuthMode
 
 from ..db import Device
@@ -13,14 +12,14 @@ from . import controller
 
 async def log_last_connection(request: Request, dev_id: str):
     device = await get_device(dev_id)
-    if config.track_device_ip:
+    if request.scope["config"].track_device_ip:
         await DeviceManager.update_last_connection(device, round(time.time()), request.client.host)
     else:
         await DeviceManager.update_last_connection(device, round(time.time()))
 
 
 async def validate_device_token(request: Request, dev_id: str):
-    if not config.device_auth.enable:
+    if not request.scope["config"].device_auth.enable:
         return
 
     # parse device token, needs to be the `TargetToken`
@@ -33,13 +32,13 @@ async def validate_device_token(request: Request, dev_id: str):
 
     device = await DeviceManager.get_device(dev_id)
     # setup mode should register devices and set up their auth token
-    if config.device_auth.mode == DeviceAuthMode.SETUP:
+    if request.scope["config"].device_auth.mode == DeviceAuthMode.SETUP:
         if device_token is None:
             return
         await DeviceManager.update_auth_token(device, device_token)
 
     # lax mode should register devices and check their token if they have one, but not register their tokens
-    elif config.device_auth.mode == DeviceAuthMode.LAX:
+    elif request.scope["config"].device_auth.mode == DeviceAuthMode.LAX:
         # should not be possible
         assert device is not None
 
@@ -47,7 +46,7 @@ async def validate_device_token(request: Request, dev_id: str):
             raise HTTPException(401, "Device authentication token does not match.")
 
     # strict mode should ensure all device are already set up and have a token, then check the token
-    elif config.device_auth.mode == DeviceAuthMode.STRICT:
+    elif request.scope["config"].device_auth.mode == DeviceAuthMode.STRICT:
         if device_token is None:
             raise HTTPException(401, "Device authentication token is required in strict mode.")
         # do not create a device in strict mode
--- a/conftest.py
+++ b/conftest.py
@@ -31,7 +31,7 @@ async def clear_cache():
     yield
 
 
-@pytest_asyncio.fixture(scope="module")
+@pytest_asyncio.fixture(scope="function")
 async def test_app():
     async with RegisterTortoise(
         app=app,
@@ -40,7 +40,7 @@ async def test_app():
         yield app
 
 
-@pytest_asyncio.fixture(scope="module")
+@pytest_asyncio.fixture(scope="function")
 async def async_client(test_app):
     async with AsyncClient(
         transport=ASGITransport(app=test_app), base_url="http://test", follow_redirects=True
@@ -116,6 +116,7 @@ async def test_data(db):
             update_mode=UpdateModeEnum.ASSIGNED,
             assigned_software=software_release,
             hardware=hardware,
+            auth_token="auth_token1",
         )
 
         yield dict(
@@ -126,4 +127,6 @@ async def test_data(db):
             rollout_default=rollout_default,
             device_rollout=device_rollout,
             device_assigned=device_assigned,
+            device_authentication=device_assigned,
+            device_no_authentication=device_rollout,
         )
--- a/tests/updater/controller/test_device_auth.py
+++ b/tests/updater/controller/test_device_auth.py
@@ -1,0 +1,151 @@
+import pytest
+
+from goosebit.settings import config
+from goosebit.settings.schema import DeviceAuthMode
+
+
+async def _api_device_get(device_auth_async_client, dev_id):
+    response = await device_auth_async_client.get("/api/v1/devices")
+    assert response.status_code == 200
+    devices = response.json()["devices"]
+    return next(d for d in devices if d["uuid"] == dev_id)
+
+
+async def _api_device_update(device_auth_async_client, device, update_attribute, update_value):
+    response = await device_auth_async_client.patch(
+        f"/ui/bff/devices",
+        json={"devices": [f"{device.uuid}"], update_attribute: update_value},
+    )
+    assert response.status_code == 200
+
+
+@pytest.mark.asyncio
+async def test_poll_strict_with_no_auth_device_with_no_auth(async_client, test_data, monkeypatch):
+    device = test_data["device_no_authentication"]
+    monkeypatch.setattr(config.device_auth, "enable", True)
+    monkeypatch.setattr(config.device_auth, "mode", DeviceAuthMode.STRICT)
+
+    response = await async_client.get(f"/ddi/controller/v1/{device.uuid}")
+    assert response.status_code == 401
+
+
+@pytest.mark.asyncio
+async def test_poll_strict_with_no_auth_device_with_auth(async_client, test_data, monkeypatch):
+    device = test_data["device_authentication"]
+    monkeypatch.setattr(config.device_auth, "enable", True)
+    monkeypatch.setattr(config.device_auth, "mode", DeviceAuthMode.STRICT)
+
+    response = await async_client.get(f"/ddi/controller/v1/{device.uuid}")
+    assert response.status_code == 401
+
+
+@pytest.mark.asyncio
+async def test_poll_strict_with_auth_device_with_auth(async_client, test_data, monkeypatch):
+    device = test_data["device_authentication"]
+    monkeypatch.setattr(config.device_auth, "enable", True)
+    monkeypatch.setattr(config.device_auth, "mode", DeviceAuthMode.STRICT)
+
+    response = await async_client.get(
+        f"/ddi/controller/v1/{device.uuid}", headers={"Authorization": f"TargetToken {device.auth_token}"}
+    )
+    assert response.status_code == 200
+
+
+@pytest.mark.asyncio
+async def test_poll_lax_with_no_auth_device_with_no_auth(async_client, test_data, monkeypatch):
+    device = test_data["device_no_authentication"]
+    monkeypatch.setattr(config.device_auth, "enable", True)
+    monkeypatch.setattr(config.device_auth, "mode", DeviceAuthMode.LAX)
+
+    response = await async_client.get(f"/ddi/controller/v1/{device.uuid}")
+    assert response.status_code == 200
+
+
+@pytest.mark.asyncio
+async def test_poll_lax_with_no_auth_device_with_auth(async_client, test_data, monkeypatch):
+    device = test_data["device_authentication"]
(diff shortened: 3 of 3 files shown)


goosebit.yaml
```diff
@@ -1,32 +1,46 @@
 ## Settings to adjust for each installation
+## Settings are listed with single #, comments with 2 #
```
easybe, line 2: All these unrelated changes should go into a separate commit. Anyway, not a fan, sorry.

goosebit.yaml
```diff
@@ -1,32 +1,46 @@
...
 
-# Port to host the server on, default:
+## Port to host the server on, default:
 # port: 60053 # GOOSE ;)
```
easybe, line 5: I suggest this convention as an alternative:
```suggestion
#port: 60053 # GOOSE ;)
```

goosebit.yaml
```diff
@@ -1,32 +1,46 @@
...
+## Whether to track the IP of the device when it polls.  Useful for debugging, but can be turned off for privacy.
 # track_device_ip: true
 
-# Secret key used for parsing user sessions. It is HIGHLY advised to pass this as an environment variable instead.
-# Defaults to a randomized value. If this value is not set, user sessions will not persist when app restarts.
+## Device authentication settings
+# device_auth:
+
+##  Enable the checking of device authentication
+#   enable: false
```
easybe, line 27: And this:
```suggestion
#  Token-based device authentication
#device_auth:
#  enable: false
```

goosebit.yaml
```diff
@@ -1,32 +1,46 @@
...
+
+##  The auth mode, one of:
+##  - strict: Strict mode, all devices must have keys, and all keys must be set via the API.
+##  - lax: Lax mode, devices which have keys must use them, but devices without keys do not require them.
+##  - setup: Setup mode, any devices polling with an auth token that don't have one will save it.
```
easybe, line 32: Is this the same as hawkBit implements it?

b-rowan: Not actually sure, but my goal is not to be hawkBit, but to be better than hawkBit.  I think this is one of the spots where we have that flexibility to change some things, but this is also why PRs exist, I am open to comment if you guys don't agree.

easybe: Well, in general, I opt for "as simple as possible" (KISS). IMHO pre-mature optimization should be avoided as much as possible. I would only implement a feature when there is an actual need for it.

I agree, we don't want to be hawkBit. IMHO "simplicity" should be gooseBit's main advantage. However, we should implement certain features (or a subset) in order to facilitate the migration to gooseBit.

b-rowan: Agreed on the simplicity side. 

The idea was that strict mode is what most people will likely want, so that is the default when unspecified.  Lax mode maybe be a ur least useful mode, but I think the one we would use if we were to use auth mode at all.  Setup mode is basically for migration only, because it makes sure there are no issues when migrate with accidentally sending something wrong to the API.

I'm ok with the only option being strict, but I figured these options are simple and useful enough that I may as well just include them.

easybe:
I think there should be some tests for this, especially because we at Husqvarna won't be using this.

b-rowan:
> I think there should be some tests for this, especially because we at Husqvarna won't be using the feature.

Biggest issue here is that it doesn't seem like we can dynamically update the config.  Might need a separate PR to attach it to the app, then it can be used via `request.app.config` or maybe `request.app.goosebit_conf`

b-rowan:
Tests are broken because it won't let me instantiate a new app for this test file (need a new one to alter the config).  No idea how to fix this.

tsagadar:
Great to have progress here. Unfortunately, could not yet look at the code - should have time on the weekend.

Conceptually I had a few thoughts
- It should be extensible to support certificate authentication as well (will be required by us)
- Cannot see the value that `lax` gives
- Ability to disable authentication is helpful to get something up and running quickly
- Devices that fail to authenticate should be logged as warnings. Alternative is to create them but assign them the state `unauthenticated`. Would simplify spotting problematic gateways but would allow to make a DoS attack by creating massive amounts of devices. A middle ground would be to limit the amount of unauthenticated devices - not sure if this is a good idea.
- `setup` seems convenient. Still not sure if it should be supported though Idea is to use this mode only temporarily. But this means that a mechanism has to be built anyways to pre-populate devices. So why not do it right from the start?

What about the following settings
- `Allow un-authenticated devices (only use in test installations)`
- `Allow device authentication by certificate (through reverse proxy)`
- `Allow device authentication by security token`

As for the behavior: if "allow un-authenticated" is enabled, still fail if reverse-proxy header / token doesn't match.

goosebit/ui/bff/devices/requests.py
```diff
@@ -10,3 +10,4 @@ class DevicesPatchRequest(BaseModel):
     pinned: bool | None = None
     feed: str | None = None
     force_update: bool | None = None
+    auth_token: str | None = None
```
tsagadar, line 13: Should it be added to the bff even though the UI does not use it? In a way it is dead code like this.

goosebit/api/v1/devices/routes.py
```diff
@@ -48,4 +48,32 @@ async def devices_delete(_: Request, config: DevicesDeleteRequest) -> StatusResp
     return StatusResponse(success=True)
 
 
+@router.patch(
+    "",
+    dependencies=[Security(validate_user_permissions, scopes=["device.write"])],
+)
+async def devices_patch(_: Request, config: DevicesPatchRequest) -> StatusResponse:
```
tsagadar, line 55: Duplicated put endpoint. PATCH IMHO should fail if resource does not exist - which is not the case.

goosebit/api/v1/devices/routes.py
```diff
@@ -48,4 +48,60 @@ async def devices_delete(_: Request, config: DevicesDeleteRequest) -> StatusResp
...
+        if config.feed is not None:
+            await updater.update_feed(config.feed)
+        if config.force_update is not None:
+            await updater.update_force_update(config.force_update)
+        if config.auth_token is not None:
+            await updater.update_auth_token(config.auth_token)
+    return StatusResponse(success=True)
+
+
+@router.put(
```
tsagadar, line 79: Add test case for this?

tsagadar: Also rather non-standard to have such a "batch" request to update/create multiple resources. But can see the advantages if one wants to modify a big number of devices.

goosebit/ui/bff/devices/routes.py
```diff
@@ -79,6 +79,8 @@ async def devices_patch(_: Request, config: DevicesPatchRequest) -> StatusRespon
             await updater.update_feed(config.feed)
         if config.force_update is not None:
             await updater.update_force_update(config.force_update)
+        if config.auth_token is not None:
```
tsagadar, line 82: Again: dead code?

goosebit/settings/schema.py
```diff
@@ -24,6 +25,17 @@ def get_json_permissions(self):
         return [str(p) for p in self.permissions]
 
 
+class DeviceAuthMode(StrEnum):
+    SETUP = "setup"  # setup mode, any devices polling with an auth token that don't have one will save it
+    STRICT = "strict"  # all devices must have keys, and all keys must be set up with the API
+    LAX = "lax"  # devices may or may not use keys, but device with keys set must poll with them
```
tsagadar, line 31: See my comment on the PR - `lax` seems pointless. What about my proposal of a different config structure?

tsagadar (changes requested):

b-rowan:
> Cannot see the value that `lax` gives

Lax mode performs exactly like `Allow un-authenticated` mode as you mention above, its just a lot shorter for having in a config.


> * `setup` seems convenient. Still not sure if it should be supported though Idea is to use this mode only temporarily. But this means that a mechanism has to be built anyways to pre-populate devices. So why not do it right from the start?

Confused what you mean here?  Setup is basically for migration from other services, if you are already using hawkbit and don't want to have to find the UUID and auth token for every device, `setup` is just to make sure there are no errors doing initial device setup.

> * `Allow un-authenticated devices (only use in test installations)`
> 
> * `Allow device authentication by certificate (through reverse proxy)`
> 
> * `Allow device authentication by security token`

This is extremely close to what I have.  Lax is allow unauthenticated, and strict is allow via token.  The only change I would make to allow the suggested configuration is add a sub-config for strict that allows you to configure the proxy headers, which is really only one option, the key used to validate the cert.  This is relatively easy to extend into, and I don't fully understand how you guys are doing it, thus I left it out.


> As for the behavior: if "allow un-authenticated" is enabled, still fail if reverse-proxy header / token doesn't match.

Lax already does this, if the token is set for a device and that token is wrong when the device makes a request, it will fail.

goosebit/api/v1/devices/routes.py
b-rowan: More worried about getting the tests to actually run and not just break when I try to instantiate a separate app, I'm sure something to do with how I'm scoping parts of the tests, but I dont really understand well enough to fix it...

goosebit/api/v1/devices/routes.py
b-rowan: Good call.  Not sure if that's in scope for this PR, can fix it here if you want.

goosebit/ui/bff/devices/requests.py
b-rowan: Maybe?  AFAIU there may be plans to show it on the UI, but fine to leave out.

goosebit/ui/bff/devices/routes.py
b-rowan: Maybe?  Should we let the user set the token via UI in the future?  If so this should be here, makes 1 less thing we will have to search for in the future, and the performance hit is negligible.

tsagadar:
> > * `setup` seems convenient. Still not sure if it should be supported though Idea is to use this mode only temporarily. But this means that a mechanism has to be built anyways to pre-populate devices. So why not do it right from the start?
> 
> Confused what you mean here? Setup is basically for migration from other services, if you are already using hawkbit and don't want to have to find the UUID and auth token for every device, `setup` is just to make sure there are no errors doing initial device

It works if one has a fix population in the market and those are all online as well. Then you can use `setup` mode for a few days and switch to strict.

Where it does not work: devices somewhere in a sales channel and can come online at any point in time (over the next few years)

Where it does not safe much effort: device still being manufactured. Would then require to create script to add newly produced device tokens to gooseBit. Having that script implemented would then also allow to load the old devices as well.

b-rowan:
> It works if one has a fix population in the market and those are all online as well. Then you can use `setup` mode for a few days and switch to strict.
> 
> Where it does not work: devices somewhere in a sales channel and can come online at any point in time (over the next few years)
> 
> Where it does not safe much effort: device still being manufactured. Would then require to create script to add newly produced device tokens to gooseBit. Having that script implemented would then also allow to load the old devices as well.

Agreed, but in most cases if you were previously using hawkbit you will have at least a few devices deployed, and will need to write that script anyway.  Setup just gives us a small advantage over other services for initial setup, so unless you think there is a big reason not to include it, I can't see any reason we wouldn't just allow it, right?

goosebit/api/v1/devices/routes.py
tsagadar: Here I used another approach to adjust the settings during the pytest execution: https://github.com/husqvarnagroup/goosebit/pull/new/mm/device_auth

goosebit/ui/bff/devices/requests.py
tsagadar: Let's add it when we get there. Otherwise: dead code.

b-rowan: Used in tests.

goosebit/ui/bff/devices/routes.py
b-rowan: Used in tests.

goosebit/api/v1/devices/routes.py
```diff
@@ -48,4 +48,60 @@ async def devices_delete(_: Request, config: DevicesDeleteRequest) -> StatusResp
     return StatusResponse(success=True)
 
 
+@router.patch(
```
tsagadar, line 51: This gets added in your first commit 1cb764b1 (message: Add auth token to device database and update migrations). It adds the duplicate for patch and put. Why do we need these two calls? Should not only be the patch code in the commit "Add auth token to API and device patch request" ?

goosebit/updater/routes.py
```diff
@@ -12,14 +12,14 @@
 
 async def log_last_connection(request: Request, dev_id: str):
     device = await get_device(dev_id)
-    if config.track_device_ip:
+    if request.scope["config"].track_device_ip:
```
tsagadar, line 15: Why do we sometimes import the config to use it and sometime we use the middleware to access it?

tsagadar (changes requested):
Few things to change. To get the changes packaged in the proper commits it can help to make a soft reset to the commit before your changes and recreate the commits one by one (at least in PyCharm it is then quite easy to select which lines should go into which commit)

goosebit/updater/routes.py
b-rowan: Middleware is slightly more dynamic, and should be used in the routes where the `Request` can be accessed.  Everywhere else can use config because it doesn't need to be dynamic or it doesn't have access to request.

goosebit/api/v1/devices/routes.py
b-rowan: Fixed, patch now raises if the device does not exist.

tsagadar (approved):

PR #197, opened 2024-11-16
Mostly refactoring

tsagadar:
A few changes - mostly things that should simplify the code base.

Creating a draft PR to get some initial feedback - best to review the different commits.


goosebit/updater/controller/v1/routes.py
```diff
@@ -34,11 +34,11 @@ async def polling(request: Request, dev_id: str, updater: DeviceUpdateManager =
...
         raise HTTPException(404)
 
-    sleep = config.poll_time_default
+    sleep = config.poll_time
 
     if device.last_state == UpdateStateEnum.UNKNOWN:
-        # device registration
-        sleep = config.poll_time_registration
+        # device registration: force device to poll again in 10s. After registration, an update might be available
+        sleep = "00:00:10"
```
b-rowan, line 35: May want to keep this configurable just in case?  I suppose easy enough to change later...

goosebit/updater/controller/v1/routes.py
```diff
@@ -22,11 +22,9 @@
 
 
 @router.get("/{dev_id}")
-async def polling(request: Request, dev_id: str, updater: DeviceManager = Depends(get_update_manager)):
+async def polling(request: Request, dev_id: str, device: Device = Depends(get_device)):
```
b-rowan, line 25: Can remove `dev_id` and use `device.uuid` here instead.

goosebit/ui/static/js/devices.js
```diff
@@ -29,6 +18,9 @@ const renderFunctions = {
         }
         return data;
     },
+    polling: (data, type) => {
+        return data ? "on time" : "overdue";
```
b-rowan, line 22: Does it make sense to do the same check as before, and otherwise just return he boolean value?  Should make filtering work...
```
        if (type === "display")
```

b-rowan:
Small comments, seems fine to me other than these.

goosebit/updater/controller/v1/routes.py
tsagadar: To me it is more of an internal decision of gooseBit to know what is a reasonable time to wait between registration and checking for an update. So don't want to hand over the burden to the user through a config. If we see a need in the future to have different settings we can bring it back.

goosebit/updater/controller/v1/routes.py
tsagadar: Fixed.

goosebit/ui/static/js/devices.js
b-rowan: > Assuming we activated client-side filtering - how would it work? User entering "True" or "False"? With the current code I'd assume that if switching to client-side filtering one could use "overdue".

I meant for sorting, I don't think we want searching.

goosebit/updater/controller/v1/routes.py
b-rowan: We may want to just move this to a constant in settings then, but this works.

b-rowan (approved):

goosebit/ui/static/js/devices.js
tsagadar: Client-side devices sorting is not really what we would activate - but it should work with my code as well, correct? Not even sure how ascending order of a boolean is implemented.

PR #216, opened 2024-11-26
Update python for tests/pages and add poetry lockfile

jameshilliard:
I think it's generally considered best practices to commit the lockfile for projects like this.


b-rowan (approved):

b-rowan:
Maybe disable the poetry actions on CI?

jameshilliard:
That's a strange error:
```
Failed to resolve 'pypi.org'
```

jameshilliard:
> Maybe disable the poetry actions on CI?

Disabled the lock action and got it working.

b-rowan:
> > Maybe disable the poetry actions on CI?
> 
> Disabled the lock action and got it working.

If it works when you push it though, then it just needs to be disabled in pre-commit-ci

PR #220, opened 2024-12-01
Making update mechanism more robust

tsagadar:
There was an edge case when deleting software/rollout before the last device reported back.

Second improvement: use API to get logs in test and don't directly access db.

### DIFF
--- a/tests/updater/controller/v1/test_routes.py
+++ b/tests/updater/controller/v1/test_routes.py
@@ -1,6 +1,6 @@
 import pytest
 
-from goosebit.db.models import Device, Hardware, Software
+from goosebit.db.models import Hardware, Software
 from goosebit.device_manager import DeviceManager, get_device
 from goosebit.settings import GooseBitSettings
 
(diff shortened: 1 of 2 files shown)


tests/updater/controller/v1/test_routes.py
```diff
@@ -283,9 +290,11 @@ async def test_up_to_date(async_client, test_data):
     await _poll(async_client, device.uuid, None, False)
 
 
-async def _assertLogLines(device, expected_line_count):
-    persisted_device = await Device.get(uuid=device.uuid)
-    log = persisted_device.last_log
+async def _assertLogLines(async_client, device, expected_line_count):
```
b-rowan, line 293: This should be named `_assert_log_lines` to keep with Python naming conventions.

b-rowan (changes requested):

tests/updater/controller/v1/test_routes.py
tsagadar: Changed.

b-rowan (approved):

PR #270, opened 2025-03-10
 Updated Selection Persistence in DataTable

aerole151:
Ensured selected rows remain selected after table refresh. Stored selected row UUIDs before reload and re-applied selection afterward. Prevents unintended row deselection after updates.


b-rowan (approved):
Works.

b-rowan (approved):

Issue #271, opened 2025-03-14
Connect to a postgres database server that requires sslmode=require and a self-signed cert.


fhuebner:
I am trying to evaluate the postgres support with a managed postgres database hosted at digitalocean. 

But that requires `ssl=true` and `sslmode=require`. They also require use self signed certs.
(https://docs.digitalocean.com/products/databases/postgresql/how-to/connect/#connection-details)

Is that somehow possible to configure that in the current master?
Drivers for Java or Node applications allow that via connection string + the ca-file in a special location on the filesystem, but  I don't find something like that for Tortoise/Python.

b-rowan:
Here's the documentation on it - https://tortoise.github.io/databases.html#passing-in-custom-ssl-certificates

We will likely need to add support for this somehow into the settings.

fhuebner:
I tried to implement it myself. It works locally, but it is failing in the docker container in the `aerich --config /aerich.toml upgrade` part of the run command.

https://github.com/unipoin/goosebit/commit/922e5dd8676a69c9f58870ad70d9763727603d23

b-rowan:
I think this is because the docker image installs gooseBit from PyPi AFAIU, you may need to alter the dockerfile.

fhuebner:
I hacked that locally, to use the self build package from `dist/goosebit.tar.gz` (build with `poetry build`)

easybe:
For local Docker build I can offer this WIP branch: https://github.com/husqvarnagroup/goosebit/commits/gardena/eb/docker_dev/

However, I think your issue is a different one.

fhuebner:
@easybe: You where right. This issue was something different.

The Database server was using a self signed cert that is not accepted by python3.13 anymore (when `ssl.VERIFY_X509_STRICT` is set, thats the new default with python3.13):
>ssl.SSLCertVerificationError: [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: Basic Constraints of CA cert not marked critical (_ssl.c:1020)

Locally my system was still using 3.12.9.

fhuebner:
close: fixed with #273

PR #273, opened 2025-03-17
Feat: Postgresql ssl self signed cert support

fhuebner:
This Pull Request adds support to connect to PostgreSQL database servers that have a self signed ca certificate and require SSL mode.

### DIFF
--- a/goosebit/db/config.py
+++ b/goosebit/db/config.py
@@ -1,3 +1,7 @@
+import logging
+from urllib.parse import parse_qs, urlparse
+
+from goosebit.db.pg_ssl_context import PostgresSSLContext
 from goosebit.settings import config
 
 TORTOISE_CONF = {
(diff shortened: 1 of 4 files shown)


fhuebner:
Completes: #271

goosebit/settings/schema.py
```diff
@@ -58,6 +58,8 @@ class GooseBitSettings(BaseSettings):
     users: list[User] = []
 
     db_uri: str = f"sqlite:///{GOOSEBIT_ROOT_DIR.joinpath('db.sqlite3')}"
+    db_ssl_crt: str = ""
```
b-rowan, line 61: May want to change this to `db_ssl_crt: Path | None = None`?

goosebit/db/config.py
```diff
@@ -8,3 +11,98 @@
...
+            logging.debug(f"verifyflag : {f} added!")
+
+    def load_verify_locations(self, file):
+        self.context.load_verify_locations(file)
+        logging.debug(f"verify-location: {file} set!")
+
+
+
+
+if config.db_ssl_crt != "":
```
b-rowan, line 71: If changed to `Path | None`, this can be `is config.db_ssl_crt is not None:`.

goosebit/db/config.py
```diff
@@ -8,3 +11,98 @@
...
+    # set certificate file
+    ssl_context.load_verify_locations(config.db_ssl_crt)
+
+    # parse and set verify-flags
+    if params and params["verifyflags"]:
+        ssl_context.parseVerifyFlags(params["verifyflags"])
+
+    # parse and set ssl verify-mode
+    if params and params["sslmode"]:
+        ssl_context.parseSSLMode(params["sslmode"])
```
b-rowan, line 100: Maybe want to use `urllib.parse` here?

```python
from urllib.parse import urlparse, parse_qs

if config.db_ssl_crt is not None:
    parsed = urlparse(config.db_uri)
    if parsed.scheme == "postgres":
        params = parse_qs(parsed.query)
        ssl_context = PostgresSSLContext()
        ssl_context.load_verify_locations(config.db_ssl_crt)
        if params.get("sslmode") is not None:
            ssl_context.parseSSLMode(params["sslmode"][0])
        if params.get("verifyFlags") is not None:
            ssl_context.parseVerifyFlags(params["verifyFlags"][0])
```

goosebit/db/config.py
```diff
@@ -8,3 +11,98 @@
...
+
+
+class PostgresSSLContext:
+    context: ssl.SSLContext
+
+    def __init__(self):
+        # create ssl context in server-auth mode: this sets verify_mode = required and check_hostname = True
+        self.context = ssl.create_default_context(purpose=ssl.Purpose.SERVER_AUTH)
+
+    def parseSSLMode(self, sslmode: str):
```
b-rowan, line 23: Is this a naming convention from the existing SSL lib?  Just want to make sure this is not something that can be changed to `parse_ssl_mode`.

b-rowan (changes requested):
Couple small changes and comments, seems good overall.

goosebit/db/config.py
fhuebner: Makes sense to use a parsing library, but then we have to do a mapping for some parameters.

goosebit/db/config.py
fhuebner: These resulted from me not correctly reading the coding style for python regarding classname/functionnames 🙄. Changed it to snake_case.

easybe:
Please squash “fixup” commits appropriately and force push, thanks.

b-rowan (approved):
Looks good.  If you could squash all commits into a single commit (something like `feature: Add support for ssl in Postgres connection`) so I can merge, as I can't squash and rebase because of the repo settings.

b-rowan:
Thanks!

Issue #280, opened 2025-04-01
Questions about the software-compatibility feature


jeff-heine:
I am having difficulty understanding the software compatibility feature of the Goosebit server and how to make it work for me.  I am using Goosebit v0.2.5.  I am not fluent in python, so I may have missed certain details of this feature.

1. Is there a way to force a ‘match to any target’ or a ‘ignore/opt out of software-compatibility’ option for an uploaded software file?
2. Do all swu files need to define a  hardware-compatibility entry in its sw-description to make them usable with the Goosebit server?
It appears that the server assigns a null value to the software-compatibility property if an swu file does not include define a  hardware-compatibility entry.  I have struggled to find a way for the server to assign a null software compatible file to any registered devices.
3. Is there any way to manually edit the software compatibility field, or change that way the server automatically defines it?

tsagadar:
@jeff-heine thanks for the great feedback. Here my take on your questions:

> Do all swu files need to define a hardware-compatibility entry in its sw-description to make them usable with the Goosebit server?
> It appears that the server assigns a null value to the software-compatibility property if an swu file does not include define a hardware-compatibility entry. I have struggled to find a way for the server to assign a null software compatible file to any registered devices.

With #287: no, software updates no longer require hardware-compatibility to be defined. gooseBit now marks such updates as compatible with "default" models / "default" revision.

> Is there a way to force a ‘match to any target’ or a ‘ignore/opt out of software-compatibility’ option for an uploaded software file?

If one wants to use gooseBit without software compatibility, then one should not include hardware-compatibility in the swu descriptor and not report hw_boardname / hw_revision to gooseBit.

> Is there any way to manually edit the software compatibility field, or change that way the server automatically defines it?

No, this is not supported.

Issue #279, opened 2025-04-01
help with getting started


jeff-heine:
I think Goosebit is an exciting project and definitely fills a gap.  As a newbie I have struggled to get something up and running.  Here are some notes that others may find useful.  I have been using Goosebit v0.2.5  with SWUpdate v2024.05.  I am not fluent in python so I might have missed the exact functionality of the server.

Although the documentation states that the server is “implementing HawkBits’s DDI API” and devices are expected “to use SWUpdate”, Goosebit has restricted support of these in the following ways.

* Goosebit DDI API supports only one tenant and that is ‘ddi’ not HawkBit's default ‘default’.  So you need to use swupdate -u “-t ddi ...”  on the target, rather than swupdate -u “-t default ...” to use the Goosebit server address http://goosebit-server-ip/ddi/controller/v1/...

*  Goosebit has a UUID property for its target devices.  This is in fact just the DDI API {controllerId} string.  In my experience this string does not need to conform to the UUID format – any value that creates a valid URI works.  For example, the swupdate -u “-i my-target ...” option generates a GET http://goosebit-server-ip/ddi/controller/v1/my-target  request to the server - this request registers a device on the server with the UUID value of ‘my-target’.  
Perhaps a les confusing label for this property in Goosebit would be ID or device-ID instead of UUID.

* Goosebit server does not support swu files that use the SWUpdate default version numbering scheme, the 4 field version major.minor.revision.build, e.g. version=”4.0.2.72”.  If you use this format for the version entry in your sw-description the server will not upload your swu file and will tell you it cannot parse your file. You need to use the alternative Semantic version format, e.g. 1.0.2
It would be really good if Goosebit could support both of SWUpdates version formats as I would like to use Goosebit with an existing project that uses the 4-digit schema.  It appears that the semver library is only used in a couple of places in the code base, so maybe not a big task, however my python knowledge is not up to suggesting code patches. 

* Goosebit has logic that links the swu software uploads with device targets using an internal 'software-compatibility' property.  This logic is based on swu file's ‘hardware-compatibility’ entry and on the target's configData hw_boardname & hw_revision values.  Unfortunately if these are not explicitly given, the implicit defaults given by the server to the swu uploads and registered device targets are not compatible with each other and do not ‘just work’.
[...]

PR #287, opened 2025-04-02
Fixing getting started issues

tsagadar:
The following commits address the issues raised in #279 / #280.

Create a draft pull request to get quick initial feedback on the implementation. Also will have to do manual testing, current changes are only validated by the unit tests.

Note: it contains a breaking change moving from `ddi` to the `default` tenant endpoint. In existing installations this can be reverted by setting the config.

### DIFF
--- a/goosebit.yaml
+++ b/goosebit.yaml
@@ -3,6 +3,9 @@
 # Port to host the server on, default:
 #port: 60053 # GOOSE ;)
 
+# Tenant for DDI endpoint
+#tenant: DEFAULT
+
 # Database to be used, default:
 #db_uri: sqlite:///<project root>/db.sqlite3
 #db_ssl_crt: /<project root>/ca-certificate.crt
--- a/goosebit/api/v1/devices/routes.py
+++ b/goosebit/api/v1/devices/routes.py
@@ -29,7 +29,7 @@ async def devices_get(_: Request) -> DevicesResponse:
     response = DevicesResponse(devices=devices)
 
     async def set_assigned_sw(d: DeviceSchema):
-        device = await get_device(d.uuid)
+        device = await get_device(d.id)
         _, target = await DeviceManager.get_update(device)
         if target is not None:
             await target.fetch_related("compatibility")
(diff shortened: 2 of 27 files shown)


goosebit.yaml
```diff
@@ -3,6 +3,9 @@
 # Port to host the server on, default:
 #port: 60053 # GOOSE ;)
 
+# Tenant for DDI endpoint
+#tenant: default
```
easybe, line 7: hawkBit actually uses `DEFAULT`, but I guess lowercase works too (which is less ugly).

b-rowan (commented):
Looks good I think, but not sure I like using `id` as a variable name since it is a built-in, maybe `device_id` is better?

goosebit.yaml
b-rowan: I think we may actually want to default to a wildcard possibly, then allow the user to specific a tenant?  Not sure on this one, but that seems like the most compatible approach.

In agreement here that `default` is less ugly, but I think it should probably be changed in some way to be compatible with `DEFAULT` since that would ensure out-of-the-box compatibility with SWUpdate.

tsagadar: Switched it to use DEFAULT for the out-of-the-box compatibility.

goosebit/db/migrations/models/4_20250402085235_rename_uuid_to_id.py
```diff
@@ -0,0 +1,11 @@
+from tortoise import BaseDBAsyncClient
+
+
+async def upgrade(db: BaseDBAsyncClient) -> str:
+    return """
+        ALTER TABLE "device" RENAME COLUMN "uuid" TO "id";"""
```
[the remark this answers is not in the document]
tsagadar: Was able to create fresh SQLite database - so seems to work.
```
% poetry run aerich upgrade                                                                                                                                                                                                                      2025-04-02 16:23:42,070 - goosebit.settings - INFO - Loading settings from: /development/gb3/goosebit.yaml
Success upgrade 0_20240830054046_init.py
Success upgrade 1_20241109151811_update.py
Success upgrade 2_20241121113728_update.py
Success upgrade 3_20241121140210_update.py
Success upgrade 4_20250402085235_rename_uuid_to_id.py
```

goosebit/api/v1/devices/routes.py
```diff
@@ -54,10 +54,10 @@ async def devices_delete(_: Request, config: DevicesDeleteRequest) -> StatusResp
     dependencies=[Security(validate_user_permissions, scopes=["device.write"])],
 )
 async def devices_patch(_: Request, config: DevicesPatchRequest) -> StatusResponse:
-    for uuid in config.devices:
-        if await Device.get_or_none(uuid=uuid) is None:
-            raise HTTPException(404, f"Device with UUID {uuid} not found")
-        device = await DeviceManager.get_device(uuid)
+    for id in config.devices:
```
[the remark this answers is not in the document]
tsagadar: Changed to dev_id

Issue #110, opened 2024-08-31
Lint commit messages with pre-commit


b-rowan:
With all the PR issues being raised about commit message format, it's likely a good idea to implement something like [commitlint](https://commitlint.js.org/#/reference-configuration) as a pre-commit hook (with something like [commitlint-pre-commit-hook](https://github.com/alessandrojcm/commitlint-pre-commit-hook).

tsagadar:
`commitlint` did not convince everybody. So closing this issue - not critical to have linting on commit messages.

PR #292, opened 2025-04-21
Add commitlint to check commit messages

tsagadar:
Implements `commitlint` as proposed in #110.

Opted to use rule set "@commitlint/config-conventional". This should implement (in my understanding) https://www.conventionalcommits.org/en/v1.0.0/ . 

Note: this differs to our previous rules in the following way
1. Mandates to use a `type` prefix
2. Mandates to write lower-case description

@easybe what do you think of https://www.conventionalcommits.org

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -64,3 +64,9 @@ repos:
     rev: v8.24.3
     hooks:
       - id: gitleaks
+  - repo: https://github.com/alessandrojcm/commitlint-pre-commit-hook
+    rev: v9.22.0
+    hooks:
+      - id: commitlint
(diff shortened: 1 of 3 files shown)


b-rowan:
Looked at the spec, seems like a good change 👍

easybe:
I am not a fan. I don't see the point of semantic commit messages unless you deduce versions from them. As stated previous, I like the style used by projects like the Linux Kernel or Git itself (https://cbea.ms/git-commit/). Look at the format used by Git for e.g. revert commits, why fight the tool on that?

tsagadar:
So let's stick to what we have. Closing PR and will close #110 as well.

PR #293, opened 2025-04-21
Login screen improvements

tsagadar:
Display error for invalid login. Also add hints for password managers (caused warning in Chrome console).

### DIFF
--- a/goosebit/ui/static/js/login.js
+++ b/goosebit/ui/static/js/login.js
@@ -1,4 +1,5 @@
-loginForm = document.getElementById("login_form");
+const loginForm = document.getElementById("login_form");
+const errorContainer = document.getElementById("login_error");
 
 async function login() {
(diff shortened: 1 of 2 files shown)


b-rowan (dismissed):
LGTM.  Only possible comment is maybe we want to use server-side validation from bootstrap directly on the form rather than appending a new element.

https://getbootstrap.com/docs/5.3/forms/validation/#server-side

b-rowan (approved):

tsagadar:
> LGTM. Only possible comment is maybe we want to use server-side validation from bootstrap directly on the form rather than appending a new element.
> 
> https://getbootstrap.com/docs/5.3/forms/validation/#server-side

With my last force push the screen just shows the server error details without knowing which input is invalid. So a form-wide alert seems preferable IMHO.

PR #323, opened 2025-07-04
feature: external service for device auth

timkambic-ngen:


goosebit/updater/routes.py
```diff
@@ -56,6 +57,27 @@ async def validate_device_token(request: Request, dev_id: str):
...
         if not device.auth_token == device_token:
             raise HTTPException(401, "Device authentication token does not match.")
+    elif request.scope["config"].device_auth.mode == DeviceAuthMode.STRICT_EXTERNAL:
+        if device_token is None:
+            raise HTTPException(401, "Device authentication token is required in strict mode.")
+        # do not create a device in strict mode
+        device = await Device.get_or_none(id=dev_id)
+        if device is None:
+            raise HTTPException(401, "Cannot register a new device in strict mode.")
+        external_request_data = {"device_id": dev_id, "secret": device_token}
```
b-rowan, line 67: This seems very specific...  Just wanted to bring it up, but is there maybe a more standard way to do this, such as like a `BasicAuth` request or something?

b-rowan (commented):
Looks good, just wanted to confirm the above first.

goosebit/updater/routes.py
jkuri: Hi @b-rowan, 

would something like this work for you:

```yaml
# goosebit.yaml

device_auth:
  enabled: true
  mode: strict_external

#  The external URL to validate tokens (to be used with "mode: strict_external")
  external_url: http://localhost:8000/auth

#  External mode, one of: basicauth, bearer, json (to be used with "mode: strict_external")
  external_mode: bearer # one of basicauth, bearer, json

#. The external JSON key to use when performing POST request (to be used with "external_mode: json")
#  external_json_key: token
```

and then in implementation;

```python
elif request.scope["config"].device_auth.mode == DeviceAuthMode.STRICT_EXTERNAL:
  if device_token is None:
      raise HTTPException(401, "Device authentication token is required in strict mode.")
  # do not create a device in strict mode
  device = await Device.get_or_none(id=dev_id)
  if device is None:
      raise HTTPException(401, "Cannot register a new device in strict mode.")
  
  # check the token with external service
  try:
      async with httpx.AsyncClient() as client:
          if request.scope["config"].device_auth.external_mode == "basicauth":
              user, pwd = device_token.split(":", 1)
              auth = httpx.BasicAuth(user, pwd)
              response = await client.post(
                  request.scope["config"].device_auth.external_url,
                  json={"token": device_token},
                  auth=auth,
              )
          elif request.scope["config"].device_auth.external_mode == "bearer":
              response = await client.post(
                  request.scope["config"].device_auth.external_url,
                  headers={"Authorization": f"Bearer {device_token}"},
              )
          elif request.scope["config"].device_auth.external_mode == "json":
              if request.scope["config"].device_auth.external_json_key is None:
                  raise HTTPException(500, "External JSON key is required for JSON mode.")
  
              json = {request.scope["config"].device_auth.external_json_key: device_token}
              response = await client.post(
                  request.scope["config"].device_auth.external_url,
                  json=json,
              )
          else:
              raise HTTPException(500, "Unsupported external mode.")
  
          if response.status_code != 200:
              raise HTTPException(401, "Device authentication token is invalid.")
  except httpx.RequestError as e:
      raise HTTPException(500, f"Error communicating with authentication service: {str(e)}")
```

Please let me know what you think and if we can improve this further.

b-rowan: The config looks good to me, the one thing I would change is we should have some validation (with pydantic's `model_validator`) that will raise these errors when starting goosebit, since it is less likely the user will see those messages when they are raised to the devices.

jkuri: I updated the PR and now working on E2E tests for this feature. 

I have a question about persisting devices into database even if they are not authenticated successfully, this is the current behavior and we didn't touch that. 
Thing is that `device = await DeviceManager.get_device(dev_id)` [here](https://github.com/timkambic-ngen/goosebit/blob/external_auth/goosebit/updater/routes.py#L35) is called every time device connects to goosebit and even if we avoid that, `log_last_connection` function calls `device = await get_device(dev_id)` [here](https://github.com/timkambic-ngen/goosebit/blob/external_auth/goosebit/updater/routes.py#L16) every single time. That means everyone have your goosebit IP and port can create as many devices in your database as he want, and they are also appearing in the UI with state unknown. I am not sure if you guys were aware of that or this is the desired behaviour?

We would like to avoid that and only have geniune device stored in goosebit database. This is easily patchable with not much code changed, but not sure why is the current behaviour as it is?

\cc @b-rowan

b-rowan (approved):

PR #329, opened 2025-08-09
test(e2e): introduce e2e tests for S3 integration

jkuri:
Hi,

I made e2e tests for S3 integration, you can try running it with:
```sh
poetry run pytest -v -s e2e/tests/test_e2e_s3.py
```

The tests include;
- creating bucket on MinIO
- uploading file to goosebit and checking if it exists on S3
- run software update on device with `swupdate` installed


e2e/Dockerfile
```diff
@@ -0,0 +1,68 @@
...
+COPY --from=builder /usr/local/bin /usr/local/bin
+
+# Set working directory
+WORKDIR /app
+
+# Copy application code
+COPY . .
+
+# Create docker-specific aerich config
+RUN printf '[tool.aerich]\ntortoise_orm = "goosebit.db.config.TORTOISE_CONF"\nlocation = "/app/goosebit/db/migrations"\n' > /aerich.toml
```
b-rowan, line 52: Why not just make a file like you're doing with goosebit.yaml and copy it?

b-rowan (approved):
Wow.  This looks amazing, I'm not even sure I'm fully qualified to review this.  Only real question I have is if it would make sense to move this to `e2e/s3/` or something like that, since this is testing with an S3 bucket instead of local filesystem storage, but that's likely unimportant.

b-rowan (approved):

PR #330, opened 2025-08-10
test(e2e): ensure tests doesn't become flaky in the future

jkuri:
Minor update for E2E tests which ensures specific version of `ubuntu` Docker image.

### DIFF
--- a/e2e/s3/swupdate/Dockerfile
+++ b/e2e/s3/swupdate/Dockerfile
@@ -52,7 +52,7 @@ RUN mkdir -p /app && apt update && apt install -y build-essential \
 
 WORKDIR /app
 
-RUN git clone https://github.com/sbabic/swupdate && cd swupdate && git reset --hard cad85ea
+RUN git clone https://github.com/sbabic/swupdate && cd swupdate && git reset --hard cad85ea38ccd53bb5174c2f5fa3fb31a8d469a39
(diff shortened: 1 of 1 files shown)


e2e/s3/swupdate/Dockerfile
```diff
@@ -52,20 +52,20 @@ RUN mkdir -p /app && apt update && apt install -y build-essential \
 
 WORKDIR /app
 
-RUN git clone https://github.com/sbabic/swupdate --depth 1
+RUN git clone https://github.com/sbabic/swupdate && cd swupdate && git reset --hard cad85ea
```
jameshilliard, line 55: Probably best to avoid short hashes here for security as they aren't collision resistant AFAIU.

jkuri: Updated to full commit hash.

e2e/s3/swupdate/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM ubuntu:latest AS build
+FROM ubuntu:plucky AS build
```
jameshilliard, line 1: This is a non-LTS release, I think noble would be better as it's the latest LTS and should have a longer support period.

jameshilliard (approved):

PR #331, opened 2025-08-13
test(e2e): improve e2e tests and reuse util functions and add additional test for device with invalid token

jkuri:


jkuri:
@easybe this PR is about adding additional test for rejecting devices with invalid token, meanwhile I updated the network configuration so it should rebase smoothy with `plugins` branch.

e2e/docker/swupdate/Dockerfile
```diff
@@ -1,78 +1,33 @@
...
-    ninja-build \
-    libzstd-dev \
-    wget \
-    python3 \
-    libebgenv-dev \
-    libmtd-dev \
-    libubi-dev \
-    libubootenv-dev \
-    libzck-dev
+RUN apk add --no-cache alpine-sdk python3 cmake zlib-dev linux-headers bsd-compat-headers yaml-dev json-c-dev curl-dev libconfig-dev \
```
easybe, line 5: I liked the previous style. Maybe in general, try to align the style with the other Dockerfiles (e.g. [Dockerfile](https://github.com/UpstreamDataInc/goosebit/blob/master/docker/Dockerfile)). However, considering this is only used for tests, I don't care that much.

e2e/docker/swupdate/entrypoint.sh
```diff
@@ -2,4 +2,4 @@
 
 openssl rand -hex 4 >/etc/device-id
 
-/usr/bin/supervisord -c /etc/supervisor/conf.d/supervisord.conf
+/usr/bin/supervisord
```
easybe, line 5: Why does one need supervisord?

jkuri: `supervisord` is used to run both `swupdate` and sample program that we are updating in tests, it's also handy to restart the services using `supervisorctl restart swupdate` without container is being stopped. In normal circumstances would be better to use `systemd` or some other alternative, but since we are running this inside `docker` I did not find a better solution.

easybe: I see. In normal circumstances, you only want one service per container. But here that is fine, of course.

b-rowan:
Once this is merged I want to review dependency updates then release v0.3.0, as I think we have had a large enough body of changes so far.  LMK thoughts.

b-rowan (changes requested):
Once this is rebased and the networks are remove this should be good to go.

jkuri:
> > networks should be there, but they are not hardcoded to specific subnets anymore.
> 
> Can't you use hostnames without networks? I think the tests were passing fine on master with them removed entirely...

you are correct, sorry I wasn't aware of that. Now they are removed.

b-rowan (approved):

PR #344, opened 2025-08-19
feat(storage): delete object from S3 when deleting via web UI

jkuri:
Currently when you delete `.swu` file via web UI it only deletes it from the database but it stays on S3 bucket. This PR fixes that, also e2e test for that was added.


tests/e2e/s3/tests/test_e2e_s3.py
```diff
@@ -217,3 +217,80 @@ def test_e2e_device_update_rollout_to_version(ensure_services_ready):
...
+                break
+            except ClientError as e:
+                last_exc = e
+                time.sleep(1.0)
+        else:
+            raise AssertionError(
+                f"Object not found in MinIO before delete bucket={MINIO_BUCKET}, key={key}: {last_exc}"
+            )
+
+        del_resp = client.request(
```
b-rowan, line 269: Should be able to use `client.delete` here for consistency.

b-rowan (changes requested):
Very minor style change.

tests/e2e/s3/tests/test_e2e_s3.py
jkuri: I am not sure why, but:

```py
del_resp = client.delete(
   "/api/v1/software",
   json={"software_ids": [sw_id]},
   headers={"Authorization": f"Bearer {token}"},
)
```

does not behave same as current `client.request`, it doesn't work. Do you have an idea why?

b-rowan: Let me look into it.

b-rowan (approved):

tests/e2e/s3/tests/test_e2e_s3.py
b-rowan: https://www.rfc-editor.org/rfc/rfc9110.html#section-9.3.5-6

This seems to explain why not.

jkuri: Otherwise for multiple deletion you would need to call endpoint multiple times, for each software id respectively.

PR #353, opened 2025-08-29
Run release CI workflow for PRs in dry run mode

easybe:
We should make sure the packages can be built before a release.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -29,15 +29,15 @@ jobs:
         uses: JRubics/poetry-publish@7100bd02517e9f82452e6247849042f6c74dde04 # v2.0
         with:
           pypi_token: ${{ secrets.PYPI_API_KEY }}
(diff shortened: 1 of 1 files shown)


b-rowan (dismissed):
Looks good to me, just need to figure out how to fix the failing jobs.

easybe:
> > @b-rowan is CI failing because you removed the secret for the PyPI token?
> 
> Nope, it's still there.

I think I know what the issue might be.

easybe:
> > > @b-rowan is CI failing because you removed the secret for the PyPI token?
> > 
> > 
> > Nope, it's still there.
> 
> I think I know what the issue might be.

Secrets are not available when running for PRs for forks.

b-rowan:
Just changed the pypi token to support all 3 of those projects, should work now outside of the PR.

b-rowan (approved):

PR #357, opened 2025-09-01
Improve PyPI metadata

easybe:


easybe:
> Only thing is I'm not sure how the releases will work with the current tagging system, since Goosebit gets released with the tagged version AFAIR right?

I am not sure I understand what you mean? The packages for the plugins are only (potentially) updated when one pushes a version tag for gooseBit. I think that is fine, as we shouldn't need to update the plugins often. In this PR's case, I update all versions (as I am touching the metadata of all packages).

Issue #365, opened 2025-09-09
UI enhancement - ability to search or filter devices based on IP address


jkuri:
It would be nice to have search filter on devices screen to filter devices also based on IPs, currently it only filters based on ID and name, I think.

PR #366, opened 2025-09-09
feature: update sorting to allow for searching and ordering by more fields

b-rowan:
Also uses database ordering for all tables directly through datatables columns.

Fixes: #365

### DIFF
--- a/goosebit/schema/devices.py
+++ b/goosebit/schema/devices.py
@@ -5,7 +5,7 @@ from datetime import datetime
 from enum import Enum, IntEnum, StrEnum
 from typing import Annotated
 
-from pydantic import BaseModel, BeforeValidator, ConfigDict, Field, computed_field
+from pydantic import BaseModel, BeforeValidator, ConfigDict, computed_field
 
 from goosebit.db.models import UpdateModeEnum, UpdateStateEnum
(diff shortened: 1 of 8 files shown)


b-rowan:
Slight reversion here, versions are now no longer sorted post database query, since this simplifies the handling of data a bit.  It does change the returned order of the software versions, but it is still deterministic, just maybe a bit strange.

jkuri (approved):
LGTM 👍

easybe:
Just making sure, has this been tested with the IP column disabled?

b-rowan:
@easybe Tested as-is, it works, but let me know if you want me to make the query filter conditional, as technically (if the database was storing the IP previously) it will still filter by the IP.

easybe:
> Slight reversion here, versions are now no longer sorted post database query, since this simplifies the handling of data a bit. It does change the returned order of the software versions, but it is still deterministic, just maybe a bit strange.

Namely, semantic versions are now sorted alphabetically. This is quite a regression :disappointed:

easybe:
> @easybe Want me to write a fix for semver sorting?

Yes, please, there was a reason this was fixed previously: d9cd3a01abbdb1565ce88a81f6b52a286c51d8d6
The commit message states:

> This is only used in the software tab of the UI so the performance penalty should be ok.

PR #367, opened 2025-09-09
bug(regression): fix software version sorting

b-rowan:


easybe:
Sorting the software table seems fine now.

Would it be possibe to do the same for the devices table? 

Screenshot from 2025-09-10 08-43-58

Note that this was broken before, but, now we can also sort by "Target Software".

b-rowan:
I think the concern with this was that there tends to be a lot more devices, so in applications with 1000s of devices I think the in-memory sorting could be an issue, but if you think it's OK I'll implement it.

easybe (approved):

PR #368, opened 2025-09-10
version: bump version number

b-rowan:


poetry.lock
```diff
@@ -1,4 +1,4 @@
-# This file is automatically @generated by Poetry 2.1.4 and should not be changed by hand.
+# This file is automatically @generated by Poetry 2.1.3 and should not be changed by hand.
```
easybe, line 1: I guess we should use the same version of Poetry...

easybe (approved):

PR #374, opened 2025-09-22
Update dependencies

jameshilliard:


b-rowan (approved):
Seems fine, E2E passes so should be fine.

PR #376, opened 2025-09-26
Enable mypy strict mode

jameshilliard:
Should hopefully prevent type issues better.


goosebit/settings/schema.py
```diff
@@ -85,7 +85,7 @@ class GooseBitSettings(BaseSettings):
 
     device_auth: DeviceAuthSettings = DeviceAuthSettings()
 
-    secret_key: Annotated[OctKey, BeforeValidator(OctKey.import_key)] = secrets.token_hex(16)
+    secret_key: OctKey = OctKey.import_key(secrets.token_hex(16))
```
b-rowan, line 88: Did you try to start the application with this?  The input type is a string AFAIU, and I'm not sure they can be converted automatically.

goosebit/updates/__init__.py
```diff
@@ -50,8 +51,8 @@ async def create_software_update(uri: str, temp_file: Path | None) -> Software:
             raise HTTPException(500, "Temporary file missing, cannot parse file information")
         filename = Path(url2pathname(unquote(parsed_uri.path))).name
 
-        dest_path = Path(update_info["hash"]).joinpath(filename)
-        uri = await storage.store_file(temp_file, dest_path)
+        dest_path = Path(update_info["hash"]) / filename
```
b-rowan, line 54: Why does this keep getting changed?  `Path.__div__` == `Path.joinpath`?  I prefer `joinpath` because it feels more verbose...

jameshilliard: idk, I went ahead and reverted to the previous style

PR #412, opened 2026-01-27
Add software version to UpdateChunk

martinhoyer:
Not sure if I'm missing something, but how else would I get the version via DDI?

### DIFF
--- a/goosebit/updates/__init__.py
+++ b/goosebit/updates/__init__.py
@@ -110,6 +110,7 @@ async def generate_chunk(request: Request, device: Device) -> list[UpdateChunk]:
     return [
         UpdateChunk(
             name=software.path.name,
+            version=software.version,
             artifacts=[
                 UpdateChunkArtifact(
                     filename=software.path.name,


b-rowan (approved):
Looks good, just want to double check the hawkbit DDI docs to see what they're doing here, then I'll merge.

goosebit/updates/__init__.py
```diff
@@ -113,6 +113,7 @@ async def generate_chunk(request: Request, device: Device) -> list[UpdateChunk]:
             artifacts=[
                 UpdateChunkArtifact(
                     filename=software.path.name,
+                    version=software.version,
```
b-rowan: Looking at the hawkbit docs, this is not defined (see https://hawkbit.eclipse.dev/#/direct-device-integration-api -> Resource for software module (Deployment Base)), it seems like this should be set one level up, under `UpdateChunk`.  Not sure why this wasn't being set before, but it should be set here - https://github.com/UpstreamDataInc/goosebit/blob/25a23c4d6d14fb546b6b73a47e9206752794a6fe/goosebit/updates/__init__.py#L111-L113

goosebit/schema/updates.py
```diff
@@ -3,6 +3,7 @@
 
 class UpdateChunkArtifact(BaseModel):
     filename: str
+    version: str
```
b-rowan: Can remove this, since it should be set a level up in `UpdateChunk`.

b-rowan (changes requested):

goosebit/updates/__init__.py
martinhoyer: Ah, my bad. That's where I wanted to put it, but was not thinking straight when I did the changes again to the fork. Thanks!

b-rowan (approved):

PR #413, opened 2026-01-28
Draft/PoC: uv, ruff, version upgrades

martinhoyer:
I bumped into trouble when committing, as pre-commit hooks were quite broken on Fedora w/ Python 3.14. I've just put together a few things more as a suggestion. 

fwiw, there are quite a few errors uncovered by additional ruff rules. I've enabled only those that are passing here. Might be worth looking into it if ruff would be adopted.


b-rowan:
Iv've also been having a ton of issues with 3.14.2, Arch based system.  Not really sure why that upgrade is breaking so much stuff, but probably related to poetry environments...

b-rowan:
Why were the migrations deleted?  Those are kindof required for backwards compatibility and upgrading existing deployments...

martinhoyer:
> Why were the migrations deleted? Those are kindof required for backwards compatibility and upgrading existing deployments...

No good reason, I just didn't have much luck with applying the migrations and wanted to see if the tests and containers are running with new version.

Issue #125, opened 2024-09-06
Document current restriction of single worker


tsagadar:
With the in-memory cache gooseBit should not be deployed with multiple workers currently.

This limitation should be documented until it is addressed (if required).

easybe:
I guess in the meantime we should configure 1 worker in the image:
https://github.com/UpstreamDataInc/goosebit/blob/5c93944236baf79f3febd1ae5f9a7bfc83aaf91d/docker/Dockerfile#L19

b-rowan:
Also pretty easy to allow a `cache' setting as a Boolean, or even allow the user to set the cache location, which we can then solve with the docker compose by using a different cache system.

PR #425, opened 2026-07-16
feat: option to disable cache and increase worker count

timkambic-ngen:
As was already discussed in https://github.com/UpstreamDataInc/goosebit/issues/125.

This PR introduces an disable toggle for Cache which enables Goosebit to not be bound to a single worker.

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,6 +57,27 @@ To use PostgreSQL, set `db_uri` or the `GOSSEBIT_DB_URI` environment variable to
 postgres://user:password@host:5432/db_name
 ```
 
+### Multiple Workers
+
+By default, gooseBit runs with a single worker process. Running multiple workers (e.g. to utilize more
+CPU cores) is supported if the following requirements are met:
+
+1. The in-memory object cache is disabled, since each worker would otherwise serve stale data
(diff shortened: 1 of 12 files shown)


timkambic-ngen:
There are still some rough edges that I am working on but the base concept is visible.
(shared cache key namespace collision, device log race, per worker OTel,...)

b-rowan:
Does it make sense to add support for a different caching backend, maybe starting with redis?  That way the caching is centralized, and if a user wants to run with multiple workers they just have to use something other than the in-memory one?

timkambic-ngen:
> Does it make sense to add support for a different caching backend, maybe starting with redis? That way the caching is centralized, and if a user wants to run with multiple workers they just have to use something other than the in-memory one?
 
Not sure caching would add anything meaningful here. For multi-worker setups, Postgres is the better choice over SQLite anyway, and it should handle the workload comfortably without a cache. 

If DB profiling ever shows it as a bottleneck, the extracted `cache.py` gives us a good starting point to introduce another backend.

I would keep this PR focused on the option to disable cache and fixing the concurrency issues (device log db race, claim update slot race).

We can open an new Issue for introducing different caching backends

PR #429, opened 2026-07-21
perf: stream s3 downloads via ranged GetObject request

timkambic-ngen:
Currently if the S3 storage backend is used. When a device requests a download chunk(64kb) from goosebit. This chunk is requested from S3 and served to  device. This is done via a single request
If device has bad connectivity this download can last longer than some proxies or s3 deployments have configured as a connection timeout. And in this case the connection is dropped. And request for next chunk will fail.

This PR changes the concept. Each chunk is requested separately via new request from S3 using ranged request. And then independently served to device.
This way it is almost impossible for s3 connection to timeout as it is very short lived.

### DIFF
--- a/goosebit/storage/s3.py
+++ b/goosebit/storage/s3.py
@@ -1,4 +1,5 @@
 import asyncio
+from functools import partial
 from typing import AsyncIterable
 from urllib.parse import urlparse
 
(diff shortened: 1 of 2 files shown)


b-rowan (approved):
Looks good, I don't understand most of this, but seems fine. Untested ACK.

Issue #428, opened 2026-07-20
Frontend issue restoring devices list


timkambic-ngen:
When one device row is selected and then tab is switched to another one and back to devices list I get this error:

`Uncaught (in promise) Error: Syntax error, unrecognized expression: unsupported pseudo: NNS1-E333619
    error https://cdn.datatables.net/v/bs5/jq-3.7.0/dt-2.3.4/b-3.2.5/r-3.0.6/sl-3.1.0/datatables.min.js:14
    PSEUDO https://cdn.datatables.net/v/bs5/jq-3.7.0/dt-2.3.4/b-3.2.5/r-3.0.6/sl-3.1.0/datatables.min.js:14
    te https://cdn.datatables.net/v/bs5/jq-3.7.0/dt-2.3.4/b-3.2.5/r-3.0.6/sl-3.1.0/datatables.min.js:14
    ne https://cdn.datatables.net/v/bs5/jq-3.7.0/dt-2.3.4/b-3.2.5/r-3.0.6/sl-3.1.0/datatables.min.js:14
    re https://cdn.datatables.net/v/bs5/jq-3.7.0/dt-2.3.4/b-3.2.5/r-3.0.6/sl-3.1.0/datatables.min.js:14
    I https://cdn.datatables.net/v/bs5/jq-3.7.0/dt-2.3.4/b-3.2.5/r-3.0.6/sl-3.1.0/datatables.min.js:14
    matches https://cdn.datatables.net/v/bs5/jq-3.7.0/dt-2.3.4/b-3.2.5/r-3.0.6/sl-3.1.0/datatables.min.js:14
    filter https://cdn.datatables.net/v/bs5/jq-3.7.0/dt-2.3.4/b-3.2.5/r-3.0.6/sl-3.1.0/datatables.min.js:14
    T https://cdn.datatables.net/v/bs5/jq-3.7.0/dt-2.3.4/b-3.2.5/r-3.0.6/sl-3.1.0/datatables.min.js:14
    filter https://cdn.datatables.net/v/bs5/jq-3.7.0/dt-2.3.4/b-3.2.5/r-3.0.6/sl-3.1.0/datatables.min.js:14
    t https://cdn.datatables.net/v/bs5/jq-3.7.0/dt-2.3.4/b-3.2.5/r-3.0.6/sl-3.1.0/datatables.min.js:20
[...]

PR #430, opened 2026-07-22
fix: don't restore selection

timkambic-ngen:
Fixes the issue described here: https://github.com/UpstreamDataInc/goosebit/issues/428

### DIFF
--- a/goosebit/ui/static/js/devices.js
+++ b/goosebit/ui/static/js/devices.js
@@ -51,6 +51,12 @@ document.addEventListener("DOMContentLoaded", async () => {
         stateSave: true,
         select: true,
         rowId: "id",
+        // Don't restore selection: Select matches row IDs as CSS selectors, and
+        // device IDs may contain invalid chars (e.g. ":") that abort table init.
+        stateLoadParams: (settings, data) => {
+            // biome-ignore lint/performance/noDelete: selection must not be restored
+            delete data.select;
+        },
         ajax: {
             url: "/ui/bff/devices",
             method: "POST",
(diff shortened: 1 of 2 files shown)


goosebit/ui/static/js/devices.js
```diff
@@ -51,6 +51,12 @@ document.addEventListener("DOMContentLoaded", async () => {
         stateSave: true,
         select: true,
         rowId: "id",
+        // Don't restore selection: Select matches row IDs as CSS selectors, and
+        // device IDs may contain invalid chars (e.g. ":") that abort table init.
+        stateLoadParams: (settings, data) => {
+            // biome-ignore lint/performance/noDelete: selection must not be restored
+            delete data.select;
+        },
```
b-rowan, line 59: Doesn't this drop the user's selection when the table gets an update (every 3 seconds)?  Any chance we can do something to handle those invalid characters (like URL encoding) rather than just deleting the selection?

goosebit/ui/static/js/settings.js
```diff
@@ -39,6 +39,12 @@ document.addEventListener("DOMContentLoaded", async () => {
         stateSave: true,
         select: true,
         rowId: "username",
+        // Don't restore selection: Select matches row IDs as CSS selectors, and
+        // usernames may contain invalid chars (".", "@", ...) that abort table init.
+        stateLoadParams: (settings, data) => {
+            // biome-ignore lint/performance/noDelete: selection must not be restored
+            delete data.select;
+        },
```
b-rowan, line 47: Same idea as above, although much less of an issue as I believe the user table doesn't get updated automatically.

b-rowan (changes requested):

